PR #6, opened 2017-08-12
Fill out the VmaAllocationInfo parameter passed to vmaAllocateX.

DustinHLand:
vmaAllocateMemory, vmaAllocateMemoryForBuffer, and vmaAllocateMemoryForImage all return before filling out the VmaAllocationInfo struct.


adam-sawicki-amd:
You are right, that was a bug. Thanks a lot!

PR #7, opened 2017-08-19
Fix memcpy and VmaVector compile errors

duodenim:
Fixes compile errors with `VmaVector` and ensures that `memcpy` is always declared


adam-sawicki-amd:
You are right. Thanks a lot!

PR #9, opened 2017-09-21
Robustify VmaStat

realitix:
Currently, VmaStat is not defined as typedef.
With GCC, this is not compiling.

Just a small fix!


adam-sawicki-amd:
Good catch, thanks a lot!

PR #13, opened 2017-09-26
Fix vulkan calls

realitix:
Can you try the demo on Windows.
I tested compilation, it's ok.
But the sample only works on Windows and I have Linux.
So can you try the sample to be sure that runtime works.

Thanks


adam-sawicki-amd:
The demo for Windows, as it is right now, works correctly with the latest code. It uses Vulkan functions imported statically - the default way.

I don't want to implement importing Vulkan functions dynamically for now because I want this sample to be as simple as possible.

realitix:
@adam-sawicki-amd Yes I know but it's the same with static since it's just at initialization time.

So it means the code works ! You can merge it.

PR #14, opened 2017-10-05
First example of xml file

realitix:
Hello @adam-sawicki-amd,

It's a work in progress.
I would like to get your opinion about it. It's how I see the xml file.

Thanks,
Jean-Sébastien

### DIFF
--- a/src/vk_mem_alloc.xml
+++ b/src/vk_mem_alloc.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<registry>
+    <!-- Command -->
+    <commands>
+        <command queues="graphics" pipeline="graphics">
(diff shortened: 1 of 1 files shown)


adam-sawicki-amd:
What format is this? Is it some standard format documented somewhere, or your own?

This library has interface in similar style as Vulkan, so XML description could also be similar I guess. Feel free to use whatever works.

Providing bindings for programming languages other than C/C++ is out of scope of this project.

realitix:
I followed the Vulkan XML file standard.
Of course providing binding is not in the scope of this project but providing a XML file may be.
It depends on you. Do you thinks it worses it ?

adam-sawicki-amd:
I'm sorry but we don't want this XML file to become part of this repository because we may not be able to maintain it. We would prefer it to be in a separate repository.

realitix:
No problem @adam-sawicki-amd

PR #17, opened 2018-01-19
Fixed some avoidable warnings when building with -Wall

cdwfs:
One was an out-of-order member in the constructor, which seems uncontroversial.

The other is an unused variable (heapIndex); is it meant to be used, or should that line be deleted entirely?


adam-sawicki-amd:
I totally agree with you. Thanks for this fix.

PR #18, opened 2018-01-24
Fixed avoidable compilation warning about unused variable.

kennyalive:


adam-sawicki-amd:
This parameter should be used. Thanks for finding this bug! I committed proper fix.

kennyalive:
Happy to hear that my small change led to a bigger finding. Thanks for the quick fix!

PR #19, opened 2018-02-27
Apple support

rextimmy:
Now that MoltenVK is released by the khronos group, we can support apple platforms. Unfortunately apple doesn't implement aligned_alloc in their cstdlib library, so i used posix_memalign instead.  clang also required a few minor changes to compile, it was just to fix a few "Call to member function 'WriteNumber' is ambiguous" etc

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -1739,8 +1739,18 @@ remove them if not needed.
 #include <mutex> // for std::mutex
 #include <atomic> // for std::atomic
 
-#if !defined(_WIN32)
+#if !defined(_WIN32) && !defined(__APPLE__)
     #include <malloc.h> // for aligned_alloc()
+#endif
+
+#if defined(__APPLE__)
+#include <cstdlib>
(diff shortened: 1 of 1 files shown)


adam-sawicki-amd:
Thank you very much for enabling the library on Apple! I have merged your pull request.

I don't like casting `size_t` to `uint64_t`, because I think it should stay 32-bit on 32-bit platforms. How is `size_t` defined on Apple? Why does it show the compiler error you mentioned? Why doesn't it just choose overloaded function for `uint32_t` or `uint64_t`, like all other compilers?

rextimmy:
Firstly no worries, happy to help. size_t on the mac is a long unsigned int (hence why it doesn't choose the overloaded function), sorry i gotta admit i didn't really think of 32bit platforms when i did the pr as macOS is 64bit only these days. If you have a better way, happy to do another PR to correct it :)

adam-sawicki-amd:
By the way, as you have Vulkan running on Apple, may I have some additional question for you regarding this platform? Please find my on Twitter as "Reg__" or any other way so we can talk.

PR #21, opened 2018-03-04
Fix undeclared identifier for Apple.

shoikeda:
Hi,
I got a compile error on macOS.
```
vk_mem_alloc.h:
1855:12: error:
      use of undeclared identifier 'VMA_NULL'
    return VMA_NULL;
```
And, I fixed the issue.
Thank you.

OS: Mac OS X 10.13.3
Compiler: clang-900.0.39.2


adam-sawicki-amd:
Thank you very much for finding and fixing this issue.

PR #28, opened 2018-05-24
Unbreak on BSDs

jbeich:
`aligned_alloc` is defined in `<stdlib.h>`even on Linux. [FreeBSD](https://github.com/freebsd/freebsd/commit/2ffad81b1757554bf9413d11af0e92c93a94575b), [DragonFly](https://github.com/DragonFlyBSD/DragonFlyBSD/commit/02b66c54cac986a0bf93435b8d5ae1b17521515b), [OpenBSD](https://github.com/openbsd/src/commit/d88f57029e5acaaaf028633c7fa15c5d7325c5cc) don't support `<malloc.h>` while GNU libc uses `<malloc.h>` for non-standard extensions: `malloc_usable_size`, `mallinfo`, `memalign`, `pvalloc`, etc.

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -2229,10 +2229,6 @@ remove them if not needed.
 #include <mutex> // for std::mutex
 #include <atomic> // for std::atomic
 
-#if !defined(_WIN32) && !defined(__APPLE__)
-    #include <malloc.h> // for aligned_alloc()
-#endif
-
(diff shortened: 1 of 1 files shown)


adam-sawicki-amd:
I submitted this change, but to "development" branch. Thanks for it! I hope that definition of aligned_malloc() function won't be missing on some platforms or compilers after this change.

PR #42, opened 2018-09-15
Suppres warnings when NOMINMAX already defined

jherico:
My project already has a CMake level definition for NOMINMAX on Win32 platforms, so this raw redefinition triggers 

`warning C4005: 'NOMINMAX': macro redefinition (compiling source file ...)`

This small change should prevent the warnings in situations like this.


rextimmy:
I originally tried to suggest this exact change when i saw the NOMINMAX define was going in, it is very common for projects to define this one themselves.

PR #53, opened 2019-01-25
Fix for "potentially uninitialized variable" warning

cdwfs:
If `allocationCount` is zero, `res` would be uninitialized here.


adam-sawicki-amd:
Good catch, thanks!

PR #55, opened 2019-02-05
Fixes VMA_STATS_STRING_ENABLED define and unused variable warning

amerkoleci:
- Correctly handle VMA_STATS_STRING_ENABLED define and undefine method not being used.
- Remove unused freeSpaceBlockSize variable.


adam-sawicki-amd:
Thank you very much for these fixes.

PR #73, opened 2019-07-13
Ensure m_VulkanFunctions pointers are initialized to null if not assigned

res2k:
In certain scenarios `m_VulkanFunctions` will be initialized with non-null garbage. (My specific case: MSVC debug build, `VMA_STATIC_VULKAN_FUNCTIONS=0`, and no `pVulkanFunctions` given.)
The garbage means that the `VMA_ASSERT`s for the functions don't trigger, rather you'll get a crash when the garbage pointer is called.
(Or worse, if the garbage happens to be callable, some possibly rather puzzling-at-first misbehaviour.)

Explicitly initializing the structure to 0 ensures that the asserts trigger as intended. (Or in release builds, prevents callable garbage pointers.)


adam-sawicki-amd:
You are right! Thank you for catching this bug. I prefer to fix it by always initializing whole structure to zeros in constructor, which I just submitted to master branch,

PR #74, opened 2019-07-25
Adding a VmaVector constructor for C++11 compatibility

benvanik:
The std::vector constructor used (count + allocator) is only available in C++14 (see: https://en.cppreference.com/w/cpp/container/vector/vector) and compiling with C++11 and `VMA_USE_STL_VECTOR` will result in compile errors. This adds a new constructor and switches to using that for compatibility.

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -4097,6 +4097,9 @@ public:
     {
     }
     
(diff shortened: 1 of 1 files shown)


adam-sawicki-amd:
Thanks a good point! Thanks for the fix.

PR #76, opened 2019-09-04
Export functions for building as a Windows standalone library

TheSpydog:
Hi there! I'm developing a Vulkan graphics backend for a C# game framework, and I would very much like to use VMA for memory management. However, I quickly found out that there is no way to directly use VMA with C# on Windows. For obvious reasons I can't include the header file in my project, and even if the library is compiled as a DLL, the functions aren't exported. So I wrote up a fix!

This PR adds a new define (`VMA_STANDALONE`) that will export all the API functions (via `__declspec(dllexport)`) with the C calling convention (`__cdecl`), wrapped into easily modifiable macros. With these changes, it's simple to build the library into a standalone DLL for consumption on Windows.

Not sure if this benefits other platforms or not, but it works great for Windows. Thanks to the code in this PR, I've been able to write a [C# binding library](https://github.com/TheSpydog/VMA-CS) and use it without issue (so far)! :smile:

### DIFF
--- a/src/VmaUsage.h
+++ b/src/VmaUsage.h
@@ -33,8 +33,8 @@
 // Uncomment to test including `vulkan.h` on your own before including VMA.
 //#include <vulkan/vulkan.h>
 
-// Uncomment to export functions when building VMA as a shared library.
-//#define VMA_DLL_EXPORT
+// Uncomment when building VMA as a standalone library.
+//#define VMA_STANDALONE
 
 /*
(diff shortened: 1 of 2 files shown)


adam-sawicki-amd:
Thank you for the contribution. I like it. I think that being able to add additional declarations around all public functions using special macro, like the `__declspec(dllexport)` on Windows, is useful for cases like yours. I will merge your pull request when I find some time for the project.

PR #87, opened 2019-11-18
Added clang compile support on windows

Nyufu:
With this change the llvm\clang-cl.exe can compile the Vma with warning level 4.


adam-sawicki-a:
Thank you for this contribution. It all looks good. I merged it.

PR #90, opened 2019-11-30
Prototype for a defragmentation interface that supports tiling optimal images

JustSid:
Hey Adam,

First of all, let me preface this by saying that this is a incomplete pull request for now! But I'm very interested in your feedback on this and to see if you are interested in having this upstream.

Essentially this is the foundational work for defragmentation that supports tiling optimal images. The basic idea is that there is a new defragmentation info `VmaDefragmentationInfo3`, which has support for callbacks that the host can fill in. These callbacks are then used to figure out what kind of allocation something is and other information, such as the create info and barrier information. The only other visible difference is that there now is an `vmaDefragmentationBegin2()` to go along with it.

My goal was to make this as little intrusive as possible, all of the old schemes still work and they work the exact same way as before. Almost none of the public interfaces have been touched, and I tried to also minimize the code changes internally.

When defragmenting, VMA still does all the same work to figure out where to move new allocations. The big difference comes at execution time (see `ApplyDefragmentationMovesGpu`). Here, VMA will interrogate the host in order to figure out where the resource is used and what image layout it is in. Then VMA will create a new image, bind it to the new memory location, and do the copying & pipeline barriers necessary to synchronize everything. Because of the way images are handled, creating and binding a new image is also no longer in the responsibility of the host application, but they are provided with another callback to let them know that a resource has changed.

For an example of this, take a look at the modified Tests.cpp which has an `TestDefragmentationImagesGpu()` function that tests the new interface.

There are a bunch of open issues that I'm left with:

- [x] No support for buffers. My thought is to provide the exact same interface as for images, and most of it is already there, except for the execution part
- [x] There is a callback to interrogate the host about the type of resource, but VMA already knows this internally from the `VmaSuballocationType`. This mainly was my idea to make this an opt-in interface, but I'm not sure if this makes for a better or worse API.
- [x] Because of the way it works, it requires a command buffer. Which is at odds with how defragmentation works on the CPU. Then again, tiling optimal images might not really be something you ever have on the CPU, so it might be okay to have this be GPU only?
- [ ] Documentation! VMA is great because it has excellent documentation, and I documented basically nothing yet.
- [ ] Bit of clean up and whitespace issues

Anyways, my question is essentially: What are your thoughts on this? Would you be interested in this upstream? And if so, what are your thoughts on the API proposal?

### DIFF
--- a/src/Tests.cpp
+++ b/src/Tests.cpp
@@ -688,6 +688,7 @@ struct AllocInfo
     VmaAllocation m_Allocation = VK_NULL_HANDLE;
     VkBuffer m_Buffer = VK_NULL_HANDLE;
     VkImage m_Image = VK_NULL_HANDLE;
+    VkImageLayout m_ImageLayout = VK_IMAGE_LAYOUT_UNDEFINED;
     uint32_t m_StartValue = 0;
     union
     {
@@ -698,6 +699,10 @@ struct AllocInfo
     void CreateBuffer(
         const VkBufferCreateInfo& bufCreateInfo,
         const VmaAllocationCreateInfo& allocCreateInfo);
+    void CreateImage(
+        const VkImageCreateInfo& imageCreateInfo,
+        const VmaAllocationCreateInfo& allocCreateInfo,
+        VkImageLayout layout);
     void Destroy();
 };
 
@@ -707,6 +712,16 @@ void AllocInfo::CreateBuffer(
 {
     m_BufferInfo = bufCreateInfo;
     VkResult res = vmaCreateBuffer(g_hAllocator, &bufCreateInfo, &allocCreateInfo, &m_Buffer, &m_Allocation, nullptr);
+    TEST(res == VK_SUCCESS);
+}
+void AllocInfo::CreateImage(
+    const VkImageCreateInfo& imageCreateInfo,
+    const VmaAllocationCreateInfo& allocCreateInfo,
+    VkImageLayout layout)
+{
+    m_ImageInfo = imageCreateInfo;
+    m_ImageLayout = layout;
+    VkResult res = vmaCreateImage(g_hAllocator, &imageCreateInfo, &allocCreateInfo, &m_Image, &m_Allocation, nullptr);
     TEST(res == VK_SUCCESS);
 }
 
@@ -904,7 +919,88 @@ static void UploadGpuData(const AllocInfo* allocInfo, size_t allocInfoCount)
         }
         else
         {
-            TEST(0 && "Images not currently supported.");
+            TEST(currAllocInfo.m_ImageInfo.format == VK_FORMAT_R8G8B8A8_UNORM && "Only RGBA8 images are currently supported.");
+            TEST(currAllocInfo.m_ImageInfo.mipLevels == 1 && "Only single mip images are currently supported.");
+
+            const VkDeviceSize size = currAllocInfo.m_ImageInfo.extent.width * currAllocInfo.m_ImageInfo.extent.height * sizeof(uint32_t);
+
+            VkBuffer stagingBuf = VK_NULL_HANDLE;
+            void* stagingBufMappedPtr = nullptr;
+            if(!stagingBufs.AcquireBuffer(size, stagingBuf, stagingBufMappedPtr))
+            {
+                TEST(cmdBufferStarted);
+                EndSingleTimeCommands();
+                stagingBufs.ReleaseAllBuffers();
+                cmdBufferStarted = false;
+
+                bool ok = stagingBufs.AcquireBuffer(size, stagingBuf, stagingBufMappedPtr);
+                TEST(ok);
+            }
+
+            // Fill staging buffer.
+            {
+                assert(size % sizeof(uint32_t) == 0);
+                uint32_t *stagingValPtr = (uint32_t *)stagingBufMappedPtr;
+                uint32_t val = currAllocInfo.m_StartValue;
+                for(size_t i = 0; i < size / sizeof(uint32_t); ++i)
+                {
+                    *stagingValPtr = val;
+                    ++stagingValPtr;
+                    ++val;
+                }
+            }
+            
+            // Issue copy command from staging buffer to destination buffer.
+            if(!cmdBufferStarted)
+            {
+                cmdBufferStarted = true;
+                BeginSingleTimeCommands();
+            }
+
+            
+            // Transfer to transfer dst layout
+            VkImageSubresourceRange subresourceRange = {
+                VK_IMAGE_ASPECT_COLOR_BIT,
+                0, VK_REMAINING_MIP_LEVELS,
+                0, VK_REMAINING_ARRAY_LAYERS
+            };
+            
+            VkImageMemoryBarrier barrier = { VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER };
+            barrier.srcAccessMask = 0;
+            barrier.dstAccessMask = 0;
+            barrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+            barrier.newLayout = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+            barrier.srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+            barrier.dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+            barrier.image = currAllocInfo.m_Image;
+            barrier.subresourceRange = subresourceRange;
+
+            vkCmdPipelineBarrier(g_hTemporaryCommandBuffer, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, 0,
+                0, nullptr,
+                0, nullptr,
+                1, &barrier);
+
+            // Copy image date
+            VkBufferImageCopy copy = {};
+            copy.bufferOffset = 0;
+            copy.bufferRowLength = 0;
+            copy.bufferImageHeight = 0;
+            copy.imageSubresource.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+            copy.imageSubresource.layerCount = 1;
+            copy.imageExtent = currAllocInfo.m_ImageInfo.extent;
+
+            vkCmdCopyBufferToImage(g_hTemporaryCommandBuffer, stagingBuf, currAllocInfo.m_Image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &copy);
+
+            // Transfer to desired layout
+            barrier.srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
+            barrier.dstAccessMask = VK_ACCESS_MEMORY_READ_BIT;
+            barrier.oldLayout = VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL;
+            barrier.newLayout = currAllocInfo.m_ImageLayout;
+
+            vkCmdPipelineBarrier(g_hTemporaryCommandBuffer, VK_PIPELINE_STAGE_TRANSFER_BIT, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, 0,
+                0, nullptr,
+                0, nullptr,
+                1, &barrier);
         }
     }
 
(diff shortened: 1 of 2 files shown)


JustSid:
Also I guess another point that is missing is not having the CI fail :P

JustSid:
I actually have some more thoughts on this... And I'd love to hear your thoughts on this as well.

I've been reviewing our requirements, and the stop the world approach to defragmenting imposes a huge burden on us. The whole goal with Vulkan is that we never have to stop, we have pre-compiled all state and know when our memory is fully specified. However, with the current way defragmenting works, it requires us to stop using all our assets, make sure the GPU is done using them, then queue the defragmentation, wait for that to complete on the GPU and retrieve our new objects. The latter can be slightly improved with the new callbacks in this pull request, where objects can become known to the host earlier and all that's needed is a pipeline barrier to sync it all up. But the other big showstopper is that while the defragmentation is running, no new objects can be made unless you put them into a new pool.

So my next stupid idea is incremental defragmentation. Essentially a multi step defragmentation process. First it'd need to generate a defragmentation plan with moves and blocks that will become empty and need deletion in the future. Those blocks would need locking so that no further allocations will be made from them, but leaving the other blocks open for new allocations (including making _new_ blocks). This would happen in a `vmaDefragmentationBegin()` equivalent with the plan saved in the defragmentation context.

Next up would be the incremental part, something like `vmaDefragmentationContinue()`. The library will create new allocations for as many moves as it possibly can before running out of space, then move them all over. It also tells the host about the new allocations so it can update whatever it needs to update (invalidate descriptor sets, create new image views, what have you). The application will then have to continue running using the new allocations, while the GPU might still be working with the old data. Once the application knows the GPU is done with all of the swapped out allocations, it will call `vmaDefragmentationContinue()` again, which will delete the old allocations that have been replaced and continue moving things around. Here it can also already delete empty memory blocks. This process would repeat until the plan finished execution. Of course, since the pools aren't locked, the execution plan needs periodic updating because of other allocations that changed.

`vmaDefragmentationContinue()` would return `VK_NOT_READY` when it needs to be invoked again, and `VK_SUCCESS` when it's done. Of course this requires multiple command buffer submissions and possible running across many frames. But the advantage is that there is no requirement to stop the world and defragment the memory. All that is required is to have a pipeline barrier after every delete/move pair, which could in theory be fine grained or just transfer stage to top of pipe.

adam-sawicki-a:
Hello,

First of all, thank you for opening this pull request. That's the biggest contribution to the library I got so far! I'm willing to upstream your code to the library and also work on it (subject to my time constraints) as decent defragmentation is a very needed feature. Right now I plan to release VMA version 2.3.0 as-is in December 2019, but this defragmentation can become a big thing for the next release.

I'm curious what's your use case. Are you developing a game or some other real-time, interactive graphics application? For which platforms? It will be great if your defragmentation implementation is field-tested in a real large scale app before it gets to the library. Designing API for it is also better approached from user's perspective, thinking about what API would be convenient to use in the app.

I analyzed your code changes and following comments come to my mind:

**Image/buffer creation** is something I would rather leave to the user. You designed a callback that queries for `VkImageCreateInfo`, calls `vkCreateImage` and `vkBindImageMemory` inside of the library. I think that's not a robust solution. Think what happens if the user wants to attach custom structures to `VkImageCreateInfo::pNext` or use `vkBindImageMemory2` and provide custom `pNext` while binding, plus lifetime management of those additional structures.

Same with **barriers**. It's hard to make them inside of VMA in a correct and robust way. Looking at current implementation, consider following issues:
- `beginImageBarrier` for newly created image has `dstAccessMask = 0`. This doesn't look correct.
- What if a resource has `VK_SHARING_MODE_EXCLUSIVE` not `CONCURRENT`, was used on graphics queue, we want to defragment on transfer queue, and so it needs queue family ownership transfer barrier? Passing `VK_QUEUE_FAMILY_IGNORED` is not correct is this case.
- `Vma*Introspection::accessMask, stageMask` - what do they represent? The last access before defragmentation? Planned next access after defragmentation? I can see it's both. What if I want the image to be accessed differently before and after defragmentation?

**Callback functions** you designed would do the job, but I would prefer to avoid such "push approach" and design something more like a "pull approach" where the user needs to call some VMA functions repeatedly, interleaved with doing his required job. That's why I also agree with you that "stop the world" approach is bad and it would be better to provide some way to defragment incrementally, using some `vmaDefragmentationContinue` function you proposed. I like that a lot.

Even better if the defragmentation process doesn't **block other allocation/free operations** on the same memory pool performed in the meantime. But this is something very difficult to implement, so I tired to avoid it as long as I could. It will require keeping some additional information in memory blocks and/or allocations to be used also during normal allocation/free. I think the approach you proposed - to designate some memory blocks to become empty after defragmentation - is the easiest and the best one. Nonetheless, it will probably require some significant refactor or rewrite of the whole defragmentation code. I'm ready to get rid of the current solution entirely in favor of this new one.

**My rough idea for the API** is that a function to be called repeatedly to perform a single defragmentation step would:
- take "defragmentation context" object, `maxBytesToMove`, `maxAllocationsToMove`
- calculate what to move and where
- reserve memory for the new copies
- fill in an user-provided array with details of allocations to be moved in that step, along with `VmaAllocation` objects.

It would be user's responsibility to:
- fetch `VmaAllocationInfo::pUserData` if needed to deduce what resource is represented by that allocation
- create new buffer/image in a specific place, bind it to the specific memory
- make sure current image is in `TRANFER_SRC` layout and the new one in `TRANSFER_DST` (or `GENERAL`), issue appropriate barrier if necessary

Then another VMA function would need to be called with given `VkCommandBuffer`. There, VMA would:
- would record `vkCmdCopy...` commands to it
- switch `VmaAllocation` objects to point to new places in memory

Then again the user would need to issue barriers that transition destination images to their desired layouts, submit this command buffer, `vkDestroyImage` old images (`VmaAllocation` objects stay the same), and make some final call to VMA when finished so that it can free their old memory.

This way, storage of `VkImageCreateInfo` etc., as well as barriers stay outside of VMA as before. What do you think about this?

Some minor comments about the current code:

`VmaAllocationType` is probably not needed as we have this information already.

I don't like `ImageAspectMaskForFormat`. How do we make sure we support all the possible formats? New ones are added with Vulkan updates. Besides, `COLOR` is not the only aspect possible when not working with depth/stencil. How about planar textures? I would try to avoid requiring this function entirely.

Returning large structures like `Vma*Introspection` from functions by value is considered a bad practice by many developers. Such data records are usually returned by output (pointer) parameter. That's how Vulkan and VMA do it as well.

JustSid:
Thank you so much for your feedback! 

As far as our use case goes, I work at Laminar Research on the Vulkan port of [https://x-plane.com](X-Plane), a professional flight simulator. We have a planet sized streaming world, that users can put additional things in such as ortho scenery to further increase memory pressure. Our two big main concerns in terms of fragmentation is the streaming of new bits of the planets (which come in as 1x1 degree chunks) and general overcommitment of memory where X-Plane has to decide at runtime to ditch higher mip levels of some textures. We are quite good at keeping within the budget right up until the moment the budget itself shrinks. In which case we have to start dumping memory, and our best bet for that is to let VMA defragment our pools. In my stop the world tests with the above code, I can easily free 500mb of memory blocks and get the driver a good chunk of memory back. However, on the other hand, our big use case for Vulkan is that we want a stutter free world. All pipelines are pre-built and all assets are accounted for before we render a frame with them, to make sure that we never have to wake up mid-frame and block anything. Right now the engine uses OpenGL, so that's been a constant problem for us. As far as platforms are concerned, we are running Vulkan and Windows and Linux (macOS uses Metal directly, not MoltenVK) and in the future maybe Android, although we have much better control over allocations there.

Re **Buffer/Image creation**: The point about `VkImageCreateInfo::pNext` is excellent (you can tell that our needs for fancy image creation is minimal). But it would indeed be much better to punt that to the user instead. 

`dstAccessMask = 0` is I _believe_ okay because the image is in undefined layout. No need to invalidate a cache when there's no data to invalidate?

With respect to queue ownership, my thought was that the resource you want to defragment needs to be owned by the queue that owns the command buffer. So no ownership transfers necessary as far as the library is concerned. Otherwise you'd have to pass in both queues and then it becomes even more of a headache because what if the queue doesn't even support the layout transitions necessary? So ideally my thought was that the queue ownership doesn't change, and you do this on a queue that supports both transfer + whatever your image currently needs. Same of course goes for buffers, but without the headache of layout transitions.

`Vma*Introspection::accessMask, stageMask` represents the queue to either wait on, so assuming a non stop-the-world defragmentation and a read only resource, this would realistically be top of pipe. Or whatever pipeline last wrote to it, so that the write doesn't run out of order with the defragment. However, it servers as a double (and maybe stupid) role, by also defining the dst stage in the reverse case. So after it's moved, this will be the pipeline stage that is unblocked from the transfer stage. So this might become an OR of two stages. Although thinking about this some more, breaking them out into src and dst might be a much better thing to avoid unnecassry GPU pipeline draining.

Re **Callback functions**: You should've seen my first idea with lambdas :P 

All kidding aside, I see where you are coming from. My idea was to make it a little bit simpler to swap out allocations, instead of having to retrieve them a second time later on. But the runtime guarantees are kinda iffy and it might make more sense to have the user bulk poll them. However, I think they are necessary for VMA to pull data, since it'd be non-ideal to have the user create a bunch of new resources that VMA might or might not use. Right now the advantage of the callbacks here is that the user only needs to provide data when VMA decided it actually wants to move the allocation. This would be even more of an advantage if resource creation is punted onto the user.


I agree with you about the lock-free-ish part of the defragmentation, I've spent the weekend trying to think of a good solution to this. I think a simple per block locking might be a somewhat good solution? When a resource allocation is requested, VMA would _try_ to take the lock for a block and otherwise move on to a different block. Of course, I can also see the big downside here where multiple cores are trying to allocate things at the same time. Maybe some kind of spinning on the lock might be an appropriate middle ground here? What I'm thinking is that if the locking is on a per block level, then during defragmentation, VMA could just lock off specific blocks that it wants to free later on. It's simple, stupid and might just work without making everything super complex? Threads could always be additionally funnelled to avoid extra block creation when multiple threads start fighting.

With your proposed API, I do like the general idea of it, but I have one thought about the barriers. My hope with moving the barriers into VMA was that they can be minimized. VMA knows what to move, so it can issue just barriers for allocations that need it and everything in place can stay there (my use case is to throw everything and the kitchen sink at VMA in terms of defragmentation and then let it figure out what to do).

If the user has to do the barriers, then the API would need to be something like this (I believe. Terribly named API ahoy):

```
VmaDefragmentationSomeResultStructure result;
vmaDefragmentationFigureOutWhatToMove(&result);

vector<MyAllocationThing> = figureOutWhatMoved(result); // List of allocations that are going to move

// Pipeline barriers here

vmaDefragmentationActuallyMove(&result);

// Other pipeline barriers here
```

So it'd be a two step process. Although the advantage would be that the user has much greater visibility into what state their resources are in at very specific times. So this might be an advantage after all? Rethinking this, I don't feel as strongly as I thought I would about where the barriers happen. Also, it'd get rid of the other callbacks necessary for resource creation, because where the first barrier happens, resource creation could also happen. And where the second barrier happens, the user can swap out their old resources for the new ones.

Lastly:

`VmaAllocationType` I agree. It's my poor mans opt in mechanism, but realistically the opt in would be actually _using_ the new API.

`ImageAspectMaskForFormat` Yyyyyyep... That one is a bit of a weird one. Punting this back onto the user really wouldn't be a stupid idea.

`Vma*Introspection` I tend to trust the compiler with regards to copy elision. I know it's not considered the nicest style for most. But I think that problem might solve itself if the user has to actually create all the resources and the callbacks go away?

adam-sawicki-a:
> I think they are necessary for VMA to pull data, since it'd be non-ideal to have the user create a bunch of new resources that VMA might or might not use. Right now the advantage of the callbacks here is that the user only needs to provide data when VMA decided it actually wants to move the allocation. This would be even more of an advantage if resource creation is punted onto the user.

I can't see why a two-step process would be needed. With the "pull" approach the library could also return for each step a request to recreate only those resources that are really going to be moved in that step. Sames goes with barriers. My idea is that this VMA call would return ("fill in an user-provided array", as I said) only the list of resources to be moved. So the variable `VmaDefragmentationSomeResultStructure result` you mentioned would already contain the list of moves to be peformed at that step, not all the resources available for defragmentation, as passed to the defragmentation begin function. Does that make sense? I'm not sure if I explained it clearly... Your code is:

```cpp
VmaDefragmentationSomeResultStructure result;
vmaDefragmentationFigureOutWhatToMove(&result);
vector<MyAllocationThing> = figureOutWhatMoved(result);
```

For me it's more like:

```cpp
const uint32_t maxAllocationsToMove = 128;
vector<VmaDefragmentationCopy> defragCopies(maxAllocationsToMove);
uint32_t defragCopyCount;

// This function:
// - calculates list of moves to be performed in this step
// - reserves memory for destinations
// - fills defragCopies array
// - may already switch VmaAllocation objects to new location
vmaDefragmentationStepBegin(defragCtx, maxAllocationsToMove, &defragCopyCount, defragCopies.data());

for(uint32_t i = 0; i < defragCopyCount; ++i)
{
    const VmaDefragmentatonCopy& copy = defragCopies[i];
    EngineTexture* myTexture = FindEngineTexture(copy.allocation);
    vkCreateImage(vkDevice, myTexture->imageCreateInfo, &myTexture->newImage);
    vkBindImageMemory(vkDevice, myTextyre->newImage, copy.dstDeviceMemory, copy.dstOffset);
    vkCmdPipelineBarrier(cmdBuf, myTexture->newImage, UNDEFINED -> TRANSFER_DST);
    vkCmdPipelineBarrier(cmdBuf, myTextyre->currImage, SAMPLED_IMAGE -> TRANSFER_SRC);
    vkCmdCopyImage(myTexture->currImage -> myTexture->newImage);
    vkCmdPipelineBarrier(cmdBuf, myTexture->newImage, TRANSFER_DST -> SAMPLED_IMAGE);
}

vkCmdSubmit(cmdBuf)
// Wait for cmdBuf to finish.

// This function frees memory of source images.
// (Maybe it could make sense to require user to pass returned data again to avoid keeping internal copy of it.)
vmaDefragmentationStepEnd(defragCtx, defragCopyCount, defragCopies.data());

for(uint32_t i = 0; i < defragCopyCount; ++i)
{
    const VmaDefragmentatonCopy& copy = defragCopies[i];
    EngineTexture* myTexture = FindEngineTexture(copy.allocation);
    vkDestroyImage(vkDevice, myTexture->currImage);
    myTexture->currImage = myTexture->newImage;
    myTexture->newImage = VK_NULL_HANDLE;
    // Recreate views and switch descriptors to point to myTexture->newTexture.
}
```

I included `vkCmdCopyImage` as an operation performed by the user as well beacuse after some thinking I concluded this is a better way to go, as this call requires information that VMA doesn't have, like texture extents.

`dstAccessMask = 0` I'm not 100% sure about that either (Vulkan barriers are hard...), but I think going from `UNDEFINED` doesn't justify specifying 0. This flag must represent the intended usage after of the resource after the barrier and must also match pipeline stage mask. Using `UNDEFINED` as source layout doesn't necesarily mean no-operation in the driver. It may initialize some metadata of an internal texture compression format that would otherwise cause GPU to crash if left garbage or not synchronized properly.

> With respect to queue ownership, my thought was that the resource you want to defragment needs to be owned by the queue that owns the command buffer. So no ownership transfers necessary as far as the library is concerned. Otherwise you'd have to pass in both queues and then it becomes even more of a headache because what if the queue doesn't even support the layout transitions necessary?

But if the resource is used on a different queue than where we want to defragment, it would be best to do the queue family ownership trasfer barrier at the same time as we transition the image to the required `TRANSFER_SRC` layout. That's why I would prefer the barriers to be done by the user.

About locking of memory blocks: Yeah, that sounds fine. If the 1st of the calls to be made in each step reserves memory for move destinations, and memory for move sources stays reserved as before, then affected memory blocks don't even need to be excluded from parallel allocation operations for the time the defragmentation step is performed. The only thing we need is to mark source block with a special flag telling that we want to make it empty so no new allocations are made out of it. The only difficulty is the question of what additional data need to be kept in `Metadata_Generic` and/or `VmaAllocation_T` during the defragmentations step.

JustSid:
I've prototyped some more on this, the old callback based system is gone. The new incremental model makes use of the defragmentation flags to signal that it's being used by the user. The basic usage would be:

	VmaDefragmentationInfo2 info = {};
	info.flags = VMA_DEFRAGMENTATION_FLAG_INCREMENTAL;
	// Fill the rest in here
	// However, no command buffer is required here!

	VkResult result = vmaDefragmentationBegin(m_allocator, &info, &stats, &context);
	if(result == VK_SUCCESS)
		return; // No work to be done

	std::vector<VmaDefragmentationStepMoveInfo> all_moves;
	all_moves.resize(...);

	result = VK_NOT_READY;

	while(result == VK_NOT_READY)
	{
		VmaDefragmentationStepInfo step_info = {};
		step_info.moveCount = all_moves.size();
		step_info.pMoves = all_moves.data();

		vmaDefragmentationStepBegin(m_allocator, &step_info, context);

		VkCommandBuffer command_buffer = ...;

		// Iterator over the moves indicated by step_info and execute them

		vkQueueSubmit(command_buffer, ...);

		// Start using the newly created resources in future command buffers!


		// Wait until the command buffer finishes executing and all
		// old resources are being stopped in use.
		// Realistically this doesn't really happen like this but is invoked
		// as result of a callback or something

		result = vmaDefragmentationStepEnd(m_allocator, context);

		// Now delete all of the previous resources.
		// VkImage, VkBuffer, VkImageView etc
	}

	vmaDefragmentationEnd(m_allocator, context);

There currently is no locking of the memory blocks, so concurrent allocations could squeeze in and prevent the blocks from being freed. Although in my tests with heavy memory pressure it worked out to be eventually consistent with what I expected to happen. In general, because `vmaDefragmentationStepBegin()` can't free the old resources yet, for a brief moment until `vmaDefragmentationStepEnd()` is called, there is heavier memory usage within the app. Although this one can't really be fixed.

I do plan on revisiting the memory block locking though, and maybe trying to see if the whole thing can be a bit cleverer. Incremental defragmentation also requires the generic defragmentation algorithm, since fast moves within a block aren't readily supported. Given the existing other constraints, that might not be an issue either though, since I don't think the fast algorithm would be used a lot.

And lastly, visualizations! I've excluded dedicated allocations, but just to show that it works in a real application. The usage before defragmenting was roughly 6.61Gb as reported by the memory budget. After it's down to 5.81Gb due to memory compacting. This was with heavy background allocations, as well as running completely on a background thread without interfering with the main thread rendering:

Usage before defrag:
egkk_memory_stats_a

Usage after defrag:
egkk_memory_stats_b

adam-sawicki-a:
Thanks for this contribution. It looks great! I will look more deeply into the code and hopefully merge it when I find more time (I have little time to dedicate to this project recently), so for now just minor comments:

"DefragmentationStep" - we discussed before that maybe it would be better to call it "DefragmentationPass", for consistency with Vulkan.

`vmaDefragmentationStepBegin` - I would rather specify `context` as second parameter and `step_info` after that.

JustSid:
Glad you are liking it! No worries with regards to time, I'm in the same boat right now with Christmas and deadlines looming.

I noticed you committed the changes in a separate commit, I was going to make this pull request mergeable once I've actually written up the documentation and fixed the locking issue and the naming from step to pass. If you want, I can do these in new pull requests (since I'm not sure this one will still merge properly?), or if you'd rather do that, you are also very welcome to do so :)

adam-sawicki-a:
I've merged your change and then made additional commit that: 1. renames "step" to "pass", 2. fixes tests, as they were not correctly deleting old `VkBuffer`/`VkImage` and switching pointer to new one in `AllocInfo`.

Again thank you very much for this significant contribution! Please feel free to propose any further changes you like. Things that come to my mind at the moment are:

- It would be good to enable call to `ValidateGpuData` in the tests for new defragmentation, to validate the contents of textures.
- I can see you've added methods `TryLock`, `TryLockWrite`, `TryLockRead`, and used one of them once. Is it really needed?
- I would like to add some parameters `maxBytesToCopy`, `maxAllocationsToCopy` to defragmentation step.
- Finally, maybe the whole API of the defragmentation feature is worth revisiting to completely remove/deprecate the old way of defragmenting using memmove. That would probably simplify the internal implementation.

PR #92, opened 2019-12-10
Use NDEBUG to detect debug build

daemyung:


daemyung:
Debug feature only works on MSVC. To support other compilers, we need to check NDEBUG is not defined on the debug build.

adam-sawicki-a:
Thank you for this contribution. I agree with you - it's better to rely on `NDEBUG` macro rather than `_DEBUG`. I fixed this in a slightly different way than you proposed.

PR #95, opened 2019-12-29
Fix C++17 shared_mutex compile error

TheLavaBlock:
In C++17 the method is called: https://en.cppreference.com/w/cpp/thread/shared_mutex/try_lock_shared


adam-sawicki-a:
Thanks for this contribution. I merged it.

Pinging @JustSid - that's your code.

PR #97, opened 2020-01-13
Fixed incorrect for loop in ProcessDefragmetnations()

JustSid:
Basically what the title says. The loops completely fails, which prevents VMA from being able to use multiple defragmentation passes to just move a subset of the data (eg. throw 10k objects at the defragmenter, and then do 10 passes with a `moveCount` of just 1000)


adam-sawicki-a:
Thank you for the fix.

PR #105, opened 2020-03-19
Add pUserData to VmaDeviceMemoryCallbacks

Didgy74:
I made this small change to the device-memory callback functionality. At present you would have to rely on global state to do anything meaningful with the callback. Adding this optional field allows the user to interpret the pointer however they wish during the callback and use custom data.

More concrete example:
I wanted to use `VK_EXT_debug_utils` object-names to name all device-allocations made by VMA as "VMA DeviceMemory #X". All my data was not stored globally but in a struct, so to do this with the current callbacks I had to create global pointers that pointed into my struct. Adding this simple field allowed me to interpret the `pUserData` as a pointer into one of my structs, no longer having to rely on global state.

This to consider when merging this PR:
Not much since this functionality is optional, but the function signatures of `PFN_vmaAllocateDeviceMemoryFunction` and `PFN_vmaAllocateDeviceMemoryFunction` have changed and won't be backwards compatible for whoever is using the callback functionality currently.

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -15060,6 +15060,7 @@ VmaAllocator_T::VmaAllocator_T(const VmaAllocatorCreateInfo* pCreateInfo) :
 
     if(pCreateInfo->pDeviceMemoryCallbacks != VMA_NULL)
     {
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for proposing this improvement. I think it's a good idea, so I merged it.

Breaking backward compatibility is fine. I bumped version number so that from now on, "master" branch is developed towards version 3.0.0.

PR #109, opened 2020-03-27
Fixed a race condition with incremental defragmentation. 

JustSid:
The issue here is that `VmaBlockVector::DefragmentationEnd()` relied on the mutex being previously locked to safely mutate its data, but with incremental defrag this isn't guaranteed to be the case anymore.

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -6523,6 +6523,7 @@ public:
         VkCommandBuffer commandBuffer);
     void DefragmentationEnd(
         class VmaBlockVectorDefragmentationContext* pCtx,
+        uint32_t flags,
         VmaDefragmentationStats* pStats);
 
     uint32_t ProcessDefragmentations(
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for this contribution. I merged it. However I'm worried about this fragment:

```cpp
    // If the mutex isn't locked we didn't do any work and there is nothing to delete.
    if(pCtx->mutexLocked || !m_hAllocator->m_UseMutex)
```
`m_UseMutex` is false when user passed `VMA_ALLOCATOR_CREATE_EXTERNALLY_SYNCHRONIZED_BIT`, meaning he uses the library from only one thread or synchronizes it externally so mutexes don't need to be used. Is your new logic still correct then?

JustSid:
I think so, but I also thought the original logic was correct :P 

The idea here is that if `m_UseMutex` is false, then `mutexLocked` is never going to be true. So the check is that if `m_UseMutex` is false, it always does the cleanup operation on the assumption that the block has been touched during defragmentation. A better solution might be to instead track whether anything was moved out of the block instead of trying to infer things based on the status of the mutex being locked or not.

PR #110, opened 2020-03-29
Exposing notPreferredFlags from vmaFindMemoryTypeIndex() in VmaAllocationCreateInfo

JustSid:
Pretty minor change all things considered. The reason that I've added this is because sometimes it might be desirable to explicitly prefer _not_ to have a specific kind of memory property. For example in our case, we don't want `HOST_COHERENT` memory ever on the current Nvidia drivers and there was no good way to achieve this previously. We don't want to lie to VMA about our usage and use that to get memory with different properties.

The Nvidia case is mostly down to the Nvidia driver taking a lock when making an allocation and holding it for the whole duration of the request. This would normally be totally okay, but it turns out that requesting write combined memory will force the NT kernel to walk the page table entry for all 256mb (with default VMA settings) of memory and setting the write combined flag. This can take up to 80ms on my system, which means the driver holds that lock for 80ms. The big problem is that other operations like submitting a command buffer to a queue _also_ take the very same lock, so we end up with the renderer pausing mid frame for 80 or so ms which is not acceptable for our real time requirements.

This is our concrete use case for this feature. Even with fixed Nvidia drivers in the future, we'll have to maintain this work around at least for some while. I'm sure there are other use cases for this feature, so I figured I might as well create a pull request for it. Obviously this breaks API in _some_ way (assuming some people don't 0 initialize the `VmaAllocationCreateInfo` struct), so you are very welcome to reject this if you prefer to stay compatible.

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -2629,6 +2629,11 @@ typedef struct VmaAllocationCreateInfo
     Set to 0 if no additional flags are prefered. \n
     If `pool` is not null, this member is ignored. */
     VkMemoryPropertyFlags preferredFlags;
+    /** \brief Flags that preferably should *not* be set in a memory type chosen for an allocation.
+
+    Set to 0 if any flags are okay. \n
+    If `pool` is not null, this member is ignored. */
+    VkMemoryPropertyFlags notPreferredFlags;
     /** \brief Bitmask containing one bit set for every memory type acceptable for this allocation.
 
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thanks for the suggestion. That's a very interesting finding about the behavior of Nvidia driver.

I'm sorry but I would prefer not to complicate the logic of memory type selection even more. When custom rules are required to select memory type, I recommend to use `VmaAllocationCreateInfo::memoryTypeBits` instead, especially when they are GPU vendor specific.

PR #113, opened 2020-04-03
Fix Clang implicit fallthrough warning.

### LINKED ISSUE #112: Implicit fallthrough warning in VmaDeviceMemoryBlock::Init when building with Clang
Error:

```
In file included from ../../third_party/VulkanMemoryAllocator/src/vk_mem_alloc.cpp:8:
../../third_party/VulkanMemoryAllocator/src/vk_mem_alloc.h(11844,9): error: fallthrough attribute is only allowed on empty statements
        VMA_FALLTHROUGH
        ^
../../third_party/VulkanMemoryAllocator/src/vk_mem_alloc.h(1998,27): note: expanded from macro 'VMA_FALLTHROUGH'
#define VMA_FALLTHROUGH [[clang::fallthrough]]
                          ^
../../third_party/VulkanMemoryAllocator/src/vk_mem_alloc.h(11844,24): note: did you forget ';'?
        VMA_FALLTHROUGH
                       ^
                       ;
1 error generated.
```

We can fix this by either removing the fallthrough or using the [[clang:fallthrough]] or C++17 [[fallthrough]] attributes:

https://clang.llvm.org/docs/AttributeReference.html#fallthrough
https://en.cppreference.com/w/cpp/language/attributes/fallthrough

null77:
This uses the [[clang:fallthrough]] tag to silence the warning in
VmaDeviceMemoryBlock::Init. Note that there is also a C++17 attribute.
However C++17 isn't available to some projects.

Fixes #112

@adam-sawicki-a  Please take a look.

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -1993,6 +1993,13 @@ available through VmaAllocatorCreateInfo::pRecordSettings.
     #define VMA_CALL_POST
 #endif
 
+// Used to silence warnings for implicit fallthrough.
+#if __has_cpp_attribute(clang::fallthrough)
+#define VMA_FALLTHROUGH [[clang::fallthrough]];
+#else
+#define VMA_FALLTHROUGH
+#endif
+
 /** \struct VmaAllocator
 \brief Represents main object of this library initialized.
 
@@ -11834,6 +11841,7 @@ void VmaDeviceMemoryBlock::Init(
     default:
         VMA_ASSERT(0);
         // Fall-through.
+        VMA_FALLTHROUGH
     case 0:
         m_pMetadata = vma_new(hAllocator, VmaBlockMetadata_Generic)(hAllocator);
     }


null77:
Does not work on Linux.

adam-sawicki-a:
The code you proposed causes a warning in Visual Studio 2017, as `__has_cpp_attribute` macro is a new C++20 feature. So we would generate a warning on one compiler by fixing another. I don't want the library to require new C++20 standard.

The library is developed on and for Visual Studio mostly and doesn't guarantee to be free from compiler warnings on all version of all compilers and platforms. See documentation chapter "Features not supported":
https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/general_considerations.html#general_considerations_features_not_supported

null77:
Yeah, the PR wasn't the right fix. I am working around this locally by applying custom warning flags. Should be easier to maintain than try to upstream fixes.

PR #115, opened 2020-04-06
Add annotations for nullability and length

### LINKED ISSUE #114: Is it possible to make the lengths of arrays a bit more explicit
I'm writing some Haskell bindings to VulkanMemoryAllocator, and it would be really nice to have information on the lengths of arrays being passed into commands in VMA, providing the same information that the `len` attribute in the Vulkan xml spec does.

For example in the below definition, perhaps we could change: `@param[out] pAllocations Pointer to array that will be filled with handles to created allocations.` to `@param[out] pAllocations Pointer to array of allocationCount VmaAllocation values that will be filled with handles to created allocations.`

```C
/** \brief General purpose memory allocation for multiple allocation objects at once.

@param allocator Allocator object.
@param pVkMemoryRequirements Memory requirements for each allocation.
@param pCreateInfo Creation parameters for each alloction.
@param allocationCount Number of allocations to make.
@param[out] pAllocations Pointer to array that will be filled with handles to created allocations.
@param[out] pAllocationInfo Optional. Pointer to array that will be filled with parameters of created allocations.

You should free the memory using vmaFreeMemory() or vmaFreeMemoryPages().

Word "pages" is just a suggestion to use this function to allocate pieces of memory needed for sparse binding.
It is just a general purpose allocation function able to make multiple allocations at once.
It may be internally optimized to be more efficient than calling vmaAllocateMemory() `allocationCount` times.

All allocations are made using same parameters. All of them are created out of the same memory pool and type.
If any allocation fails, all allocations already made within this function call are also freed, so that when
returned result is not `VK_SUCCESS`, `pAllocation` array is always entirely filled with `VK_NULL_HANDLE`.
*/
VMA_CALL_PRE VkResult VMA_CALL_POST vmaAllocateMemoryPages(
    VmaAllocator allocator,
    const VkMemoryRequirements* pVkMemoryRequirements,
    const VmaAllocationCreateInfo* pCreateInfo,
    size_t allocationCount,
    VmaAllocation* pAllocations,
    VmaAllocationInfo* pAllocationInfo);
```

Putting this in the documentation would not be the end of the world, but better yet would be in an attribute on the member/parameter like `__attribute__ ((length (allocationCount)))`, this may be a little too *out there* though. This would be behind a macro like `LEN_ANN(allocationCount)` to avoid tripping `-Wattributes`.

Optionality information I *think* is complete in the documentation already, but I'm not certain.

expipiplus1:
These are only present on the declarations in the C API.

New macros
- `VMA_NULLABLE`
- `VMA_NOT_NULL`
- `VMA_NULLABLE_NON_DISPATCHABLE`
- `VMA_NOT_NULL_NON_DISPATCHABLE`
- `VMA_LEN_IF_NOT_NULL`

On clang `VMA_NOT_NULL` and `VMA_NULLABLE` are defined as `_Nonnull` and
`_Nullable`. Clang will use these annotations to generate diagnostics
when these functions or structs are incorrectly used.

`VMA_NULLABLE_NON_DISPATCHABLE` and `VMA_NOT_NULL_NON_DISPATCHABLE` are
not defined on platforms where Vulkan's non-dispatchable handles are not
pointers; on platforms where they are, these macros are defined to be
`VMA_NULLABLE` and `VMA_NOT_NULL` respectively.

`LEN_IF_NOT_NULL` is not defined for any compiler. It is included to
mirror the `len` attribute in the Vulkan XML spec in order to aid in
development of other language bindings for this library.

It was tempting to use C's array syntax (`foo(int len, int arr[len])`)to
specify lengths, however this is not possible to use with optional
pointers, nor is it possible to use inside structs. For the sake of
consistency only `VMA_LEN_IF_NOT_NULL` is used (additionally neither
clang not gcc generate warnings for function misuse when the array
length specifier is anything but an integer literal).

- Also correct a couple of typos
- Also add assert on ppName in vmaGetPoolName

Closes #114

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -1999,6 +1999,11 @@ available through VmaAllocatorCreateInfo::pRecordSettings.
 // The length may be one of
 // - The name of another parameter in the argument list where the pointer is declared
 // - The name of another member in the struct where the pointer is declared
+// - The name of a member of a struct type, meaning the value of that member in
+//   the context of the call. For example
+//   VMA_LEN_IF_NOT_NULL("VkPhysicalDeviceMemoryProperties::memoryHeapCount"),
+//   this means the number of memory heaps available in the device associated
+//   with the VmaAllocator being dealt with.
 #ifndef VMA_LEN_IF_NOT_NULL
     #define VMA_LEN_IF_NOT_NULL(len)
 #endif
@@ -2304,7 +2309,7 @@ typedef struct VmaAllocatorCreateInfo
     blocks to system RAM. This driver behavior can also be controlled using
     VK_AMD_memory_overallocation_behavior extension.
     */
-    const VkDeviceSize* VMA_NULLABLE pHeapSizeLimit;
+    const VkDeviceSize* VMA_NULLABLE VMA_LEN_IF_NOT_NULL("VkPhysicalDeviceMemoryProperties::memoryHeapCount") pHeapSizeLimit;
 
     /** \brief Pointers to Vulkan functions. Can be null.
 
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -3423,7 +3423,7 @@ typedef struct VmaDefragmentationInfo2 {
     It is safe to pass allocations that are in the lost state - they are ignored.
     All allocations not present in this array are considered non-moveable during this defragmentation.
     */
-    VmaAllocation VMA_NOT_NULL * VMA_NOT_NULL VMA_LEN_IF_NOT_NULL(allocationCount) pAllocations;
+    const VmaAllocation VMA_NOT_NULL * VMA_NOT_NULL VMA_LEN_IF_NOT_NULL(allocationCount) pAllocations;
     /** \brief Optional, output. Pointer to array that will be filled with information whether the allocation at certain index has been changed during defragmentation.
 
     The array should have `allocationCount` elements.
@@ -3448,7 +3448,7 @@ typedef struct VmaDefragmentationInfo2 {
     Using this array is equivalent to specifying all allocations from the pools in `pAllocations`.
     It might be more efficient.
     */
-    VmaPool VMA_NOT_NULL * VMA_NULLABLE VMA_LEN_IF_NOT_NULL(poolCount) pPools;
+    const VmaPool VMA_NOT_NULL * VMA_NULLABLE VMA_LEN_IF_NOT_NULL(poolCount) pPools;
     /** \brief Maximum total numbers of bytes that can be copied while moving allocations to different places using transfers on CPU side, like `memcpy()`, `memmove()`.
     
     `VK_WHOLE_SIZE` means no limit.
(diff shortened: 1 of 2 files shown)


src/vk_mem_alloc.h
```diff
@@ -3415,12 +3472,12 @@ typedef struct VmaDefragmentationInfo2 {
...
     */
-    VkCommandBuffer commandBuffer;
+    VkCommandBuffer VMA_NULLABLE commandBuffer;
 } VmaDefragmentationInfo2;
 
 typedef struct VmaDefragmentationPassMoveInfo {
-    VmaAllocation allocation;
-    VkDeviceMemory memory;
+    VmaAllocation VMA_NOT_NULL allocation;
+    VkDeviceMemory VMA_NOT_NULL_NON_DISPATCHABLE memory;
```
expipiplus1, line 3480: Are the annotations here correct?

src/vk_mem_alloc.h
```diff
@@ -3430,7 +3487,7 @@ To be used with function vmaBeginDefragmentationPass().
 */
 typedef struct VmaDefragmentationPassInfo {
     uint32_t moveCount;
-    VmaDefragmentationPassMoveInfo* pMoves;
+    VmaDefragmentationPassMoveInfo* VMA_NOT_NULL pMoves VMA_LEN_IF_NOT_NULL(moveCount);
```
expipiplus1, line 3490: Is the nullability annotation here correct?

expipiplus1:
Thanks! This is excellent! There are some preliminary haskell bindings generated using these annotations here: https://github.com/expipiplus1/vulkan/blob/vma/VulkanMemoryAllocator/src/Graphics/VulkanMemoryAllocator.hs 

I'll upload these to Hackage after giving them a test drive.

expipiplus1:
@adam-sawicki-a I just noticed that I had included the rejected `Add const qualifier to handles where possible` commit in this PR. You might want to revert that if that's still considered rejected! Sorry!

adam-sawicki-a:
Thanks for this quality contribution!

Including #117 is fine - I decided I will eventually merge it.

Should `VMA_LEN_IF_NOT_NULL` have its parameter as string "" or not? You mixed this in your code.

expipiplus1:
Yeah, it can be any expression and I couldn't think of a better way of saying `VkPhysicalDeviceMemoryProperties::memoryHeapCount`. Perhaps the comment here could be improved https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/115/commits/08bc59bee51b3cc8eb11a43595ec17e66f1346b2#diff-44540aefde48c442bbf9b6a5c0fc3fdeR2002 to mention say that in this case it is passed as a string to be parsed by whatever program is consuming this information.

Another way of putting it would be `VMA_LEN_IF_NOT_NULL(VkPhysicalDeviceMemoryProperties.memoryHeapCount)` but I think that might obscure the meaning.

adam-sawicki-a:
Let me know when your Haskell binding is finished or in Beta stage and I will add it to the list in VMA README.md.

adam-sawicki-a:
I made some fixes in your code. Please take a look and see if they are correct (e.g. `VMA_NOT_NULL` -> `VMA_NOT_NULL_NON_DISPATCHABLE` for `VkInstance`).

expipiplus1:
Thanks for double checking! I made a few comments.

adam-sawicki-a:
OK, fixed. Thanks again!

PR #116, opened 2020-04-06
Fix a typo

Ybalrid:
I think this title was supposed to be "Binaries"


adam-sawicki-a:
That's right, thanks!

PR #117, opened 2020-04-07
Add const qualifier to handles where possible

expipiplus1:
This mirrors the Vulkan specification where handles are often passed with a const qualifier


adam-sawicki-a:
I'm sorry but I don't think this is a good idea.

1. Vulkan API doesn't use `const` very often for object handles, mostly for structures.
2. It would break backward compatibility for the interface of this library.
3. It's not semantically correct to consider allocation objects `const` during defragmentation as they are certainly changed during this process.

expipiplus1:
The below writing came across as quite antagonistic and it's hard to correct in the medium of text, that isn't my intention at all! :) Thank you for the super library!

A little bit about the motivation for my change too: At the moment these pointers to non-const handles look a lot like "out" parameters, when their purpose is strictly input.

>     1. Vulkan API doesn't use `const` very often for object handles, mostly for structures.

I should have clarified in my commit message that this is only touching handles which are passed indirectly.

The Vulkan API never, aside from handle creation, passes handles to commands indirectly without `const`. The 16 commands which have parameters of pointers to handles always have the handle specified with `const`. For structs there are no instances of indirect handles without `const` but 20 instances with `const`. (I think this is all true, but it's not inconceivable that my script to find these examples could be missing a a couple of cases)

>     2. It would break backward compatibility for the interface of this library.

Although this would break peoples code if they are doing something like the below, 
isn't `foo(const int* handles)` callable everywhere that `foo(int* handles)` is? 

```C
  VkResult (*myVmaDefragment)(
    VmaAllocator,
    VmaAllocation*,
    VmaAllocation*,
    size_t,
    VkBool32*,
    const VmaDefragmentationInfo*,
    VmaDefragmentationStats*) = vmaDefragment;
```

>     3. It's not semantically correct to consider allocation objects `const` during defragmentation as they are certainly changed during this process.

The objects they point to are changed, but not the handles themselves, for example `vkResetFences` takes `const VkFence* pFences` as a parameter, yet the fence object is changed.

Sadly Vulkan doesn't have anything like `#define VK_DEFINE_CONST_HANDLE(object) typedef struct object##_T const * object##_const;` which is what would really be useful here.

adam-sawicki-a:
OK, you are right! Thanks for this contribution and for your explanation. I merged it.

PR #118, opened 2020-04-07
Add CPP error if trying to use recording on a non-Windows platform

expipiplus1:


adam-sawicki-a:
Thanks for this contribution. I think this error message is a good idea.

The message after #error doesn't need to be in quotes.

expipiplus1:
Thanks!

PR #126, opened 2020-04-13
Port memory recording in VMA to C++11.

IAmNotHanni:
Hello.

> - Current implementation of recording in VMA, as well as VmaReplay application, is
  coded and tested only on Windows. Inclusion of recording code is driven by
  `VMA_RECORDING_ENABLED` macro. Support for other platforms should be easy to
  add. Contributions are welcomed.

* Use C++11 to get thread id.
* Use C++11 to get time.
* Find alternative to sprintf_s..
* Use errno_t and fopen_s on Windows, fopen on other systems.

Code is tested on Windows and Linux but still needs reviews.

I am a huge fan of Vulkan memory allocator and I'm happy to contribute to it.
I'm also a huge fan of your talk "Porting your engine to Vulkan or DX12" btw.

Thank you,
Johannes.


IAmNotHanni:
I got this to compile on Linux using gcc in Travis: https://travis-ci.org/github/inexorgame/vulkan-renderer.
Still extensive feedback on Linux is necessary.

adam-sawicki-a:
Thank you for this contribution. It looks good. However, `std::stringstream` is known to be extremely slow and hated by most game developers, and so is `std::string`, so using them while processing every single VMA command doesn't look too good. Can you replace it with old good `sprintf`, or something like this, so the change doesn't worsen the performance comparing to the existing code for Windows?

IAmNotHanni:
I will search a solution for this.

IAmNotHanni:
I created this benchmark:

```
#include <benchmark/benchmark.h>
#include <Windows.h>
#include <thread>
#include <sstream>

static void GetThreadUsingStringStream(benchmark::State& state)
{
	for(auto _ : state)
	{
		auto thread_id = std::this_thread::get_id();
		std::stringstream thread_id_converter;
		thread_id_converter << std::this_thread::get_id();
		uint64_t id = static_cast<uint64_t>(std::stoull(thread_id_converter.str()));
	}
}

static void GetThreadUsingPointer(benchmark::State& state)
{
	for(auto _ : state)
	{
		auto id = std::this_thread::get_id();
		void* thread_id = &id;
		uint64_t* thread_id_ptr = static_cast<uint64_t*>(thread_id);
		// cout << *thread_id_ptr << endl;
	}
}

static void GetThreadIdUsingWindowsMethod(benchmark::State& state)
{
	for(auto _ : state)
	{
		auto thread_id = GetCurrentThreadId();
	}
}

BENCHMARK(GetThreadUsingStringStream);
BENCHMARK(GetThreadUsingPointer);
BENCHMARK(GetThreadIdUsingWindowsMethod);

BENCHMARK_MAIN();
```

Results: (release)

get_thread_id_results

I am now sure how safe this approach is however. Since ``std::this_thread::get_id()` is actually an ostream, it could be of any data size the system is using. I found it quite hard to convert or cast the data otherwise.

Alternatively, we could use the sstream solution only for other platforms than windows with conditional compiling.

What do you think?

adam-sawicki-a:
OK, doing `#ifdef` for Windows with more efficient implementation and `#else` with more generic one sounds good.

src/vk_mem_alloc.h
```diff
@@ -14629,7 +14628,7 @@ VkResult VmaRecorder::Init(const VmaRecordSettings& settings, bool useMutex)
     m_UseMutex = useMutex;
     m_Flags = settings.flags;
 
-#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
+#if defined(_WIN32)
```
IAmNotHanni, line 14631: I simplified check for windows platform here since it is used like this everywhere in the code.

src/vk_mem_alloc.h
```diff
@@ -15106,9 +15105,8 @@ VmaRecorder::UserDataString::UserDataString(VmaAllocationCreateFlags allocFlags,
         else
         {
             // If VMA_ALLOCATION_CREATE_USER_DATA_COPY_STRING_BIT is not specified, convert the string's memory address to a std::string and store it.
-            std::stringstream user_data_address_to_string_converter;
-            user_data_address_to_string_converter << pUserData;
-            m_Str = user_data_address_to_string_converter.str();
+			snprintf(m_PtrStr, 17, "%p", pUserData);
```
IAmNotHanni, line 15108: This is C++11's standard version of sprintf_s.

turol: You should use `sizeof(m_PtrStr)` instead of 17 so it automatically stays in sync if someone changes it

adam-sawicki-a: Let's not over-generalize. I don't expect the size of a pointer to go beyond 64 bits in the following years, at least on typical PC and mobile platforms.

adam-sawicki-a:
I would like to merge your pull request, but GitHub shows conflicts. It looks like the entire file is one big conflict. Did you change line endings?

src/vk_mem_alloc.h
```diff
@@ -7326,19 +7329,18 @@ class VmaRecorder
     {
     public:
         UserDataString(VmaAllocationCreateFlags allocFlags, const void* pUserData);
-        const char* GetString() const { return m_Str; }
+        const char* GetString() const { return m_Str.c_str(); }
 
     private:
         char m_PtrStr[17];
-        const char* m_Str;
+        std::string m_Str;
```
adam-sawicki-a, line 7336: Do we really need to use `std::string` even on Windows? This class is very slow, it does dynamic memory allocation.

IAmNotHanni:
I think by comitting I accidentally changed line endings. Should I just open a new PR based on the current master ?

src/vk_mem_alloc.h
IAmNotHanni: I will change it back.

adam-sawicki-a:
Possibly you can make a new commit to your PR changing line endings.

IAmNotHanni:
Is the original file UTF-8 encoded and uses Windows (CR+LF) for line endings?

turol:
The file appears to have been committed to git with windows line endings. This is generally considered a Bad Idea since it leads to exactly this kind of conflict. The right way to fix it is to convert it to unix line endings and configure your windows git client to convert automatically on checkout/commit.

IAmNotHanni:
This seems to have fixed line ending issues.

adam-sawicki-a:
I merged your change by hand. I also needed to fix order of included files as you included C++ header inside `extern "C"`, which failed to compile in Visual Studio 2017.

adam-sawicki-a:
Thanks for this great contribution.

PR #136, opened 2020-07-03
Correct if in mapping example

GreenFoxLight:
We want to know if the memory is mappable, which means that the memory properties should include the `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT`.


adam-sawicki-a:
Thank you for this fix!

PR #141, opened 2020-07-28
Fix a typo in the explanation of VMA_ALLOCATION_CREATE_MAPPED_BIT

dkorkmazturk:
I think the documents are generated from the source files so the modification to docs/html/vk__mem__alloc_8h.html may not be necessary. However I wasn't sure so just added the change to this commit anyway.


adam-sawicki-a:
Thank you for this fix.

PR #146, opened 2020-08-21
Fix misspellings.

AshkanAliabadi:
Our internal linter caught these.


src/vk_mem_alloc.h
```diff
@@ -3329,7 +3329,7 @@ VMA_CALL_PRE VkResult VMA_CALL_POST vmaResizeAllocation(
 
 /** \brief Returns current information about specified allocation and atomically marks it as used in current frame.
 
-Current paramters of given allocation are returned in `pAllocationInfo`.
+Current paramteres of given allocation are returned in `pAllocationInfo`.
```
turol, line 3332: This is also wrong, it should be "parameters"

adam-sawicki-a:

src/vk_mem_alloc.h
AshkanAliabadi: The irony. :) Sorry I didn't get to fix this in time. :/

PR #153, opened 2020-10-28
Make binary search overflow safe

r-barnes:
Our internal static analysis flags potentially unsafe midpoint
calculations. The proposed fix ensures the binary search doesn't
overflow at the cost of a single additional operation.


adam-sawicki-a:
Thank you for this contribution and for running static analysis on the code.

PR #154, opened 2020-10-28
Drop trailing whitespace

r-barnes:
The only change this commit makes is to drop trailing whitespace.
This makes contributions easier for folks whose editors are set up
to automatically remove such whitespace on save.


adam-sawicki-a:

Please note that the editor I use for development (Visual Studio 2017) adds these whitespaces for indentation automatically, so I can't promise they won't appear in the new code.

I also think that while working with an external library, the editor should be configured to not modify lines that you don't intent to modify, whether we talk about reformatting or removing trailing spaces. Doing it makes comparison and merging difficult.

r-barnes:
Thanks @adam-sawicki-a. I separated out my substantive changes for this reason.

PR #158, opened 2020-12-03
Refer to std string and stream with their fully qualified names.

AshkanAliabadi:
Hey Adam!  I keep running into build errors when I turn on debugging because of this.  Cheers.


adam-sawicki-a:
You are right. Thank you for this contribution.

PR #163, opened 2021-02-15
mem_alloc: Optimize allocation conflict checking

kd-11:
This is a port of an optimization I did for rpcs3 here: https://github.com/RPCS3/rpcs3/commit/5a049d41fd06fb4f75cfb0d955cdbe8f0c5d921e
Basically, avoids scanning suballocations when there is no need to. When allocAlignment and bufferImageGranularity are the same, a lot of time can be saved when managing thousands of small allocations frequently.
1. When aligning resulting offset, the alignment is already done beforehand to the allocAlignment. The whole reason to check for conflicts is to align the result to the bufferImageGranularity as a worst case fallback, but this is clearly not needed when allocAlignment = bufferImageGranurality.
2. When checking for conflicts after allocation, there is no need to check suballocations if the 'page' is perfectly aligned start and end.

Combining this with always using allocAlignment equal to bufferImageGranurality when making allocation requests significantly speeds up performance on non-AMD hardware, where bufferImageGranurality is typically not 1 (saved 50ms in one awful scene for example). As long as bufferImageGranurality is not expected to change during runtime, this works nicely. If I missed some corner case, let me know and we can just close this. I'd just rather not maintain separate patches if possible :p

### DIFF
--- a/src/vk_mem_alloc.h
+++ b/src/vk_mem_alloc.h
@@ -9656,7 +9656,7 @@ bool VmaBlockMetadata_Generic::CheckAllocation(
 
         // Check previous suballocations for BufferImageGranularity conflicts.
         // Make bigger alignment if necessary.
-        if(bufferImageGranularity > 1)
+        if(bufferImageGranularity > 1 && bufferImageGranularity != allocAlignment)
         {
             bool bufferImageGranularityConflict = false;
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for this contribution! It looks good. I merged it.

Issue #168, opened 2021-05-25
Add CMake support


IAmNotHanni:
Hello. Adding CMake would make it easier to work with VMA on Linux (and Windows in my opinion). Furthermore CMake could create project maps for VMA library, the sample application, VMA replay, and even the shaders. 

I am aware not everyone likes CMake, however I must point out that even [Microsoft's STL repository](https://github.com/microsoft/STL) switched from their own build system to CMake. (More information on that in the following talk: [C++20 STL Features: One Year of Development on GitHub - Stephan T. Lavavej - CppCon 2020](https://youtu.be/8kjRx8vo6y4?t=222))

If you don't mind I'm workint on a pull request for this. (I also will work on https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/130 soon).

adam-sawicki-a:
I agree with you. Cmake is a standard these days. Switching from Premake to Cmake is on my TODO list for this project.

IAmNotHanni:
Alright. I'm working on it :)

IAmNotHanni:
I'm finished. Before I open he pull request: Should it be only one `CMakeLists.txt` or should it be split up into the subfolders?

adam-sawicki-a:
Your contribution is welcome, please feel free to propose this pull request.

PR #171, opened 2021-05-27
Add CMake support

IAmNotHanni:
Closes https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/168

## Targets

In addition to the `VulkanMemoryAllocator` library target there are the following optional targets:

| Target | Description | CMake option | Default setting |
| ------------- | ------------- | ------------- | ------------- |
| VmaSample | VMA sample (see screenshot below) | `VMA_BUILD_SAMPLE` | `OFF`
| VmaSampleShaders | Shaders for VmaSample | `VMA_BUILD_SAMPLE_SHADERS` | `OFF`
| VmaReplay | Replay tool for VMA .csv trace files | `VMA_BUILD_REPLAY` | `OFF`

Please note that while VulkanMemoryAllocator library is supported on other platforms besides Windows, VmaSample and VmaReplay are not.

## CMake options (updated)

| CMake option | Description | Default setting |
| ------------- | ------------- | ------------- |
| `VMA_RECORDING_ENABLED` | Enable VMA memory recording for debugging | `OFF` |
| `VMA_USE_STL_CONTAINERS` | Use C++ STL containers instead of VMA's containers | `OFF` |
| `VMA_STATIC_VULKAN_FUNCTIONS` | Link statically with Vulkan API | `OFF` (updated) |
| `VMA_DYNAMIC_VULKAN_FUNCTIONS` | Fetch pointers to Vulkan functions internally (no static linking) | `ON` (updated) |
| `VMA_DEBUG_ALWAYS_DEDICATED_MEMORY` | Every allocation will have its own memory block | `OFF` |
| `VMA_DEBUG_INITIALIZE_ALLOCATIONS` | Automatically fill new allocations and destroyed allocations with some bit pattern | `OFF` |
| `VMA_DEBUG_GLOBAL_MUTEX` | Enable single mutex protecting all entry calls to the library | `OFF` |
| `VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT` | Never exceed [VkPhysicalDeviceLimits::maxMemoryAllocationCount](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#limits-maxMemoryAllocationCount) and return error | `OFF` |

## Changelog
- [x] Use `find_package(Vulkan REQUIRED)`
- [x] Make sure MSVC uses Unicode
- [x] Make sure C++14 is used
- [x] Make sure MSVC uses multiple cores for compiling
- [x] Output .spv files to /bin folder
- [x] Find glslangValidator and compile shaders with it
- [x] Add optional setup for VmaReplay on Windows
- [x] Add optional setup for VmaSample on Windows
- [x] Add an info message that VmaSample and VmaReplay are not supported on Linux

## TODO
- [x] Check which macros in the code can be made CMake options
- [x] Add build instructions for Windows and Linux
- [x] Test if shader compilation works on Linux
- [x] Fix `VMA_BUILD_EXAMPLE_APP_SHADERS` being ignored on Linux
- [x] Fix VmaReplayApp build on Windows

## Screenshot (updated)
updated_screenshot

### Outlook
In the future, maybe somebody can do this:
* Test if this CMake setup works on Android as well
* Add build instructions for MacOS and Android

### Acknowledgments
Thanks to [IceflowRE](https://github.com/IceflowRE) and [yeetari](https://github.com/yeetari). This pull request is inspired by their work.
Furthermore I greet everybody from Vulkan Discord :)


pezcode:
I think it would make sense to hide the example and the replay tool behind options, something like `VMA_BUILD_EXAMPLE`. Most people probably don't want these built, and/or don't have glslangValidator in PATH.

IAmNotHanni:
We can make it optional:

```cmake
option(VMA_BUILD_EXAMPLE_APP "Build example application" OFF)
option(VMA_BUILD_REPLAY_APP "Build VMA replay application" OFF)
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    if("${FILE}" MATCHES ".*\\.cpp")
+        set(GROUP "Source Files\\${GROUP}")
+    elseif("${FILE}" MATCHES ".*\\.hpp")
+        set(GROUP "Header Files\\${GROUP}")
+    endif()
+
+    source_group("${GROUP}" FILES "${FILE}")
+endforeach()
+
+add_library(VmaLibrary ${VMA_LIBRARY_SOURCE_FILES})
```
DethRaid, line 23: Why name it `VmaLibrary`? I think that `VulkanMemoryAllocator` would be a better name imo

src/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
+add_subdirectory(Shaders)
+add_subdirectory(VmaReplay)
```
DethRaid, line 2: This always builds `VmaReplay`. As a consumer of VMA, I don't want to build the replay application with my application. Please give me a CMake option to disable building both this and the shaders

src/CMakeLists.txt
DethRaid, line 23: Why name it `VmaLibrary`? I think that `VulkanMemoryAllocator` would be a better name

CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.9)
+
+project(VulkanMemoryAllocator)
+
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
+
+find_package(Vulkan REQUIRED)
+
+# Put binaries into bin folder
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/bin)
```
DethRaid, line 10: Will this affect the build of my project?

src/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+        set(GROUP "Source Files\\${GROUP}")
+    elseif("${FILE}" MATCHES ".*\\.hpp")
+        set(GROUP "Header Files\\${GROUP}")
+    endif()
+
+    source_group("${GROUP}" FILES "${FILE}")
+endforeach()
+
+IF (WIN32)
+    add_executable(VmaExample ${VMA_EXAMPLE_SOURCE_FILES})
```
DethRaid, line 71: Same as `VmaReplay` - give me an option to disable building this target

DethRaid (changes requested):

CMakeLists.txt
IAmNotHanni: On Windows this makes Visual Studio output the binary into the `bin` folder. Usually the binary belongs to the `build` folder which is excluded from the git repository.

DethRaid: Sure... but if my application uses VMA, and VMA's CMakeLists.txt sets the runtime output directory, will that affect what directory Visual Studio outputs my own application to? The point is that I don't want it to

src/CMakeLists.txt
IAmNotHanni: Done.

CMakeLists.txt
IAmNotHanni: Oh that is a good point I didn't think of this! I only added it because the same directory also contains the compiled shaders. I will remove this setting and just change the working directory of the visual studio debugger. Since that's the only platform which supports VmaExample, that should do.

src/CMakeLists.txt
IAmNotHanni: Done.

src/CMakeLists.txt
IAmNotHanni: Done.

CMakeLists.txt
IAmNotHanni: Setting the output directory to `bin` is also bad because this means it's not in the gitignore.

DethRaid: Yeah...

IAmNotHanni: Done.

expipiplus1:
These are really question for vmausage.cpp, nevertheless: Does this compile
cleanly with clang?

Is it possible to add options for things like VMA_STATIC_VULKAN_FUNCTIONS
and VMA_DEBUG?

Here is what I've been using for some time for the haskell bindings
https://github.com/expipiplus1/vulkan/blob/master/VulkanMemoryAllocator/src/lib.cpp

IAmNotHanni:
I successfully tested Debug and Release build using `gcc 9.3.0` and I can also compile it using `0.0.0-4ubuntu1` on Ubuntu 20.04 LTS:

`cc=clang cxx=clang++ cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Debug`
`cc=clang cxx=clang++ cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Release`

@expipiplus1 That's a good idea in my opinion. Furthermore we also could add more options to the CMake settings instead of embedding it somewhere into the code as switchable macro. One thing I liked to do in my projects is to set `VMA_RECORDING_ENABLED` through CMake.

IAmNotHanni:
- [ ] Test if this CMake setup works on Android as well

IAmNotHanni:
I will read through the code and look for possible options to add to CMake. Exposing them in the CMake UI makes it clearer to the users that they exist.

IAmNotHanni:
@expipiplus1 Here's a logfile when using

`cc=clang cxx=clang++ cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-Wall -Wextra"`

[logfile_clang.log](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/files/6560774/logfile_clang.log)

IAmNotHanni:
- [ ] Oh I comitted my spirv binaries by accident. I will edit and force push that commit.

IAmNotHanni:
We could expose the following macros as CMake settings:

| Option | Default value | Description |
| ------------- | ------------- | ------------- |
`VMA_RECORDING_ENABLED` | 0 |  While using the library, sequence of calls to its functions together with their parameters can be recorded to a file and later replayed using standalone player application.
`VMA_USE_STL_CONTAINERS` | 0 | Define this macro to 1 to make the library use STL containers instead of its own implementation.
`VMA_STATIC_VULKAN_FUNCTIONS` | 1 | By default, VMA assumes you you link statically with Vulkan API. If this is not the case, `#define VMA_STATIC_VULKAN_FUNCTIONS 0` before `#include` of the VMA implementation and use another way.
`VMA_DYNAMIC_VULKAN_FUNCTIONS` | 0 | You can `#define VMA_DYNAMIC_VULKAN_FUNCTIONS 1` and make sure `vkGetInstanceProcAddr` and `vkGetDeviceProcAddr` globals are defined. All the remaining Vulkan functions will be fetched automatically.
`VMA_DEBUG_ALWAYS_DEDICATED_MEMORY` | 0 | Every allocation will have its own memory block. Define to 1 for debugging purposes only.
`VMA_DEBUG_ALIGNMENT` | 1 | Minimum alignment of all allocations, in bytes. Set to more than 1 for debugging purposes only. Must be power of two.
`VMA_DEBUG_MARGIN` | 0 | Minimum margin before and after every allocation, in bytes. Set nonzero for debugging purposes only.
`VMA_DEBUG_INITIALIZE_ALLOCATIONS` | 0 | Define this macro to 1 to automatically fill new allocations and destroyed allocations with some bit pattern.
`VMA_DEBUG_DETECT_CORRUPTION` | 0 | Define this macro to 1 together with non-zero value of VMA_DEBUG_MARGIN to enable writing magic value to the margin before and after every allocation and validating it, so that memory corruptions (out-of-bounds writes) are detected.
`VMA_DEBUG_GLOBAL_MUTEX` | 0 | Set this to 1 for debugging purposes only, to enable single mutex protecting all entry calls to the library. Can be useful for debugging multithreading issues.
`VMA_DEBUG_MIN_BUFFER_IMAGE_GRANULARITY` | 1 | Minimum value for VkPhysicalDeviceLimits::bufferImageGranularity. Set to more than 1 for debugging purposes only. Must be power of two.
`VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT` | 0 | Set this to 1 to make VMA never exceed VkPhysicalDeviceLimits::maxMemoryAllocationCount and return error instead of leaving up to Vulkan implementation what to do in such cases.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+set_target_properties(
+    VulkanMemoryAllocator PROPERTIES
+
+    CXX_EXTENSIONS OFF
+    # Use C++14
+    CXX_STANDARD 14
+    CXX_STANDARD_REQUIRED ON
+)
+
+target_include_directories(
```
charles-lunarg, line 39: Any particular reason why `target_include_directories` has all its contents on separate lines? (purely style nitpick)
Feel free to ignore this, I just felt it used a lot of space and I don't think it'd be harder to read if it was condensed to one line.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.9)
+
+project(VulkanMemoryAllocator)
+
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
```
charles-lunarg, line 5: 1. Would this affect projects which do insource builds? While convenient for safeties sake, its good to not enforce something if not necessary. If it doesn't propagate *out* of the VMA CMakeLists.txt file, this isn't an issue.
2. Should this go before or after the `project()` call?
3. These are undocumented features (that I only learned about today). Anything that is intentionally undocumented make me wary of using them, as I have no idea how to determine what they do without either playing with it extensively or asking someone who does know.
4. This doesn't prevent a CMakeCache.txt or CMakeFiles/ folders from being created either.

charles-lunarg (commented):
There's a lot of small to medium sized questions or comments.
I'm not a package maintainer so i'm not super familiar with the intricacies for installation, so I can't comment on the packagability of this.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+find_package(Vulkan REQUIRED)
```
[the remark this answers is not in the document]
adam-sawicki-a: I don't have a strong opinion on this, but I think requiring Vulkan SDK to be installed when using VMA is fine.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+
+        target_link_libraries(
+            VmaExample
+
+            PRIVATE
+            VulkanMemoryAllocator
+            Vulkan::Vulkan
+        )
+    else()
+        message(STATUS "VmaExample application is not supported to Linux")
```
adam-sawicki-a, line 114: Current state of the project is to compile the sample app, with code of the actual app disabled using `#ifdef _WIN32` inside source files, so that other platforms at least check if the library code compiles. As I understand, in your Cmake the VMA is still compiled because it is a separate LIB, so we won't check less than before by disabling the sample app?

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+    elseif("${FILE}" MATCHES ".*\\.hpp")
+        set(GROUP "Header Files\\${GROUP}")
+    endif()
+
+    source_group("${GROUP}" FILES "${FILE}")
+endforeach()
+
+if (VMA_BUILD_EXAMPLE_APP)
+    if(WIN32)
+        add_executable(VmaExample ${VMA_EXAMPLE_SOURCE_FILES})
```
adam-sawicki-a, line 77: Actual source file is called "VulkanSample.cpp", so I think we should stick to one convention - either call the project VmaSample or rename the file to *Example.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+if(VMA_BUILD_REPLAY_APP)
+    add_subdirectory(VmaReplay)
+endif()
+
+set(VMA_LIBRARY_SOURCE_FILES
+    VmaUsage.cpp
+)
+
+# Use folder structure as filters in project map
+foreach(FILE ${VMA_LIBRARY_SOURCE_FILES})
```
adam-sawicki-a, line 14: Is this grouping into source and header files really needed, or just a purely visual grouping for Visual Studio? If the latter, maybe we could drop this, as there are only few files in each project, while this would make CMakeLists.txt files shorter by a large %.

src/CMakeLists.txt
adam-sawicki-a: There is also inconsistency in naming - VmaReplayApp and VmaExampleAppShaders end with "App" while VmaExample doesn't. I suggest to remove the "App" part.

adam-sawicki-a:
Thank you for all the work on this. I checked it and it works fine. Please let me know when you consider it ready and I will merge the pull request.

Controlling macros like `VMA_STATIC_VULKAN_FUNCTIONS` from Cmake is a good idea as long as they can still be used the old way. Many game developers prefer to just include "vk_mem_alloc.h" in their projects rather than downloading an external repository from GitHub as a submodule, compile the library as a separate LIB etc.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+endforeach()
+
+if (VMA_BUILD_EXAMPLE_APP)
+    if(WIN32)
+        add_executable(VmaExample ${VMA_EXAMPLE_SOURCE_FILES})
+
+        # Visual Studio specific settings
+        if(${CMAKE_GENERATOR} MATCHES "Visual Studio.*")
+            # Use Unicode instead of multibyte set
+            add_definitions(-DUNICODE -D_UNICODE)
```
adam-sawicki-a, line 82: This command has been superseded by newer add_compile_definitions.

CMakeLists.txt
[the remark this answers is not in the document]
charles-lunarg: While I may disagree about `find_package(Vulkan REQUIRED)`, specifically just the REQUIRED part, I am fine with this requirement. I would like to explain why I made the comment in the first place though.

It is derived from my experience with PR's made to `vk-bootstrap` for support in the conan package manager. The issue with `find_package` was that conan has its own `Vulkan-Headers` package and `find_package` doesn't know where these headers are located. Instead, the packaging script had to modify the CMakelists.txt to remove `find_package`, replacing it with conan-specific stuff.
 
I know that VMA is very popular library and already has been added to package managers like vcpkg, so I hope that this cmake file makes it easier to add to package managers, well the ones that use cmake. I am by no means experienced enough with packaging to know what is required and isn't, so maybe there's a better way than what I did for vk-bootstrap.

For reference, this PR discusses why/how the original version of vk-bootstrap required modifying the CMakelists.txt.  (scroll to the bottom to get past the build notifications)  https://github.com/conan-io/conan-center-index/pull/4738

src/CMakeLists.txt
IAmNotHanni: It's not needed. I will remove it.

IAmNotHanni:
Ok so I will go through the comments by @charles-lunarg and @adam-sawicki-a.
After this, I will introduce the options proposed by @expipiplus1. Then we should review it again.

IAmNotHanni:
@adam-sawicki-a do you know somebody who can test this for Android and MacOS?

README.md
```diff
@@ -94,6 +94,22 @@ With this one function call:
 
 `VmaAllocation` is an object that represents memory assigned to this buffer. It can be queried for parameters like Vulkan memory handle and offset.
 
+# How to build
+
+On Windows it is recommended to use [CMake UI](https://cmake.org/runningcmake/). Alternatively you can generate a Visual Studio project map using CMake in command line: `cmake -B./build/ -DCMAKE_BUILD_TYPE=Debug -G "Visual Studio 16 2019" -A x64 ./`
+
+On Linux, use CMake with [Ninja](https://ninja-build.org/) and run `cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Debug`
```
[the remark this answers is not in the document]
IAmNotHanni: I replaced it with:

```
mkdir build
cd build
cmake ..
make
```

README.md
```diff
@@ -94,6 +94,22 @@ With this one function call:
...
+
+The following CMake options are available
+
+| Target | Description | CMake option |
+| ------------- | ------------- | ------------- |
+| VmaExample | VMA example application | `VMA_BUILD_EXAMPLE_APP` |
```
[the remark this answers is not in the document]
IAmNotHanni: Done.

src/CMakeLists.txt
IAmNotHanni: Done.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+
+set(VMA_EXAMPLE_SOURCE_FILES
+    Common.cpp
+    SparseBindingTest.cpp
+    Tests.cpp
+    VulkanSample.cpp
+)
+
+# Use folder structure as filters in project map
+foreach(FILE ${VMA_EXAMPLE_SOURCE_FILES})
```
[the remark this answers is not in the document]
IAmNotHanni: The `source_group` commands have been removed.

IAmNotHanni:
Oh and also: is squashing the commits desired before merging it?

README.md
```diff
@@ -94,6 +94,22 @@ With this one function call:
...
+On Windows it is recommended to use [CMake UI](https://cmake.org/runningcmake/). Alternatively you can generate a Visual Studio project map using CMake in command line: `cmake -B./build/ -DCMAKE_BUILD_TYPE=Debug -G "Visual Studio 16 2019" -A x64 ./`
+
+On Linux, use CMake with [Ninja](https://ninja-build.org/) and run `cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Debug`
+
+The following CMake options are available
+
+| Target | Description | CMake option |
+| ------------- | ------------- | ------------- |
+| VmaExample | VMA example application | `VMA_BUILD_EXAMPLE_APP` | 
+| VmaShaders | Shaders for VmaExample | `VMA_BUILD_EXAMPLE_APP_SHADERS` |
```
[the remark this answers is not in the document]
IAmNotHanni: Done.

CMakeLists.txt
IAmNotHanni: Since I could not find clear documentation about whether this command propagates or not, I removed it.

src/CMakeLists.txt
IAmNotHanni: I made it into one line. Maybe there is some `clang-format` like tool for CMake which could be used in the future to keep everything consistent.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+)
+
+target_link_libraries(
+    VulkanMemoryAllocator
+
+    PUBLIC
+    Vulkan::Vulkan
+)
+
+set(VMA_EXAMPLE_SOURCE_FILES
```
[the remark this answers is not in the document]
IAmNotHanni: Right. I moved it into the 

```
if (VMA_BUILD_EXAMPLE_APP)
    if(WIN32)
```

block below it.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
+if(VMA_BUILD_EXAMPLE_APP_SHADERS)
```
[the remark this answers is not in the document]
IAmNotHanni: That's right. Since it is less important I moved it to the end of that file.

src/CMakeLists.txt
IAmNotHanni: I added this because of the following reason: On Linux the code looks like this:

```
#else // #ifdef _WIN32

#include "VmaUsage.h"

int main()
{
}

#endif // #ifdef _WIN32
```

On some Linux platforms like my Ubuntu 20.04 compiling the VMA example application does **not** result in an executable file, but it generates a shared object instead:

```
file ./bin/VmaExample
./bin/VmaExample: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=ffa8949669582487287a59f98e8f472e0acf854e, for GNU/Linux 3.2.0, with debug_info, not stripped
```

I think the compiler sees that the application does nothing and prefers to make it a shared object then? I can't explain why this is the case, but it took me quite some time to understand what's wrong. I added this to inform other Linux devs. Some others tested it for me and on their system an executable (which does nothing) was obtained. I think in both cases it's better to inform the developers that the example application is not ported to Linux.

src/CMakeLists.txt
IAmNotHanni: Done.

src/CMakeLists.txt
IAmNotHanni: I removed the "App" part everywhere, renamed "VmaExample" to "VmaSample" and made it consistent.

CMakeLists.txt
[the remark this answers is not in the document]
IAmNotHanni: I will keep it as it is for now. I think that's probably a pull request for somebody with more experience than me.

src/CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+)
+
+target_include_directories(
+    VulkanMemoryAllocator
+
+    PUBLIC
+    ${PROJECT_SOURCE_DIR}/include
+)
+
+target_link_libraries(
```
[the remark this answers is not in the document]
IAmNotHanni: I will continue to work on this tomorrow.

adam-sawicki-a:
I checked the latest code - it looks and works well :)

Unfortunately I don't know anyone who would test it on Android or MacOS. CI builds for the project are also setup only for Windows and Linux. Let's do it for Windows and Linux and wait for external contributions to fix other platforms if necessary.

Squashing is not needed. I also commit small changes to the repository.

src/CMakeLists.txt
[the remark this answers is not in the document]
IAmNotHanni: I think the only way to fix this is to expose `VMA_DYNAMIC_VULKAN_FUNCTIONS` as CMake option. Then we can check if it's defined and then we don't need to link against `Vulkan::Vulkan`. Also I think we should only make `VMA_DYNAMIC_VULKAN_FUNCTIONS` an option and not `VMA_STATIC_VULKAN_FUNCTIONS`, because they are mutually exclusive and then we don't need to check if both are defined by accident.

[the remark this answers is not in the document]
IAmNotHanni: Done.

IAmNotHanni:
Ok I'm done and from my side it's ready to be merged. I exposed the following VMA settings macros as CMake options:

| CMake option | Description | Default setting |
| ------------- | ------------- | ------------- |
| `VMA_RECORDING_ENABLED` | Enable VMA memory recording for debugging | `OFF` |
| `VMA_USE_STL_CONTAINERS` | Use C++ STL containers instead of VMA's containers | `OFF` |
| `VMA_DYNAMIC_VULKAN_FUNCTIONS` | Fetch pointers to Vulkan functions internally (no static linking) | `OFF` |
| `VMA_DEBUG_ALWAYS_DEDICATED_MEMORY` | Every allocation will have its own memory block | `OFF` |
| `VMA_DEBUG_INITIALIZE_ALLOCATIONS` | Automatically fill new allocations and destroyed allocations with some bit pattern | `OFF` |
| `VMA_DEBUG_GLOBAL_MUTEX` | Enable single mutex protecting all entry calls to the library | `OFF` |
| `VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT` | Never exceed [VkPhysicalDeviceLimits::maxMemoryAllocationCount](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#limits-maxMemoryAllocationCount) and return error | `OFF` |

The CMake options will be passed on as follows:

Settings

IAmNotHanni:
Just as a quick note for the future: since VMA contains a lot of code parts which are compiled only if the macros are enabled, there are code parts which are not covered by the CI. For example if I enable `VMA_RECORDING_ENABLED`, I get the following gcc warnings: [gcc-warnings.log](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/files/6580946/gcc-warnings.log) There are only warning.

However, enabling `VMA_USE_STL_CONTAINERS` will result in errors: [gcc-errors-use-STL-containers.log](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/files/6580953/gcc-errors-use-STL-containers.log)

I will look into this in issue https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/175

src/VmaReplay/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+if (WIN32)
+    set(VMA_REPLAY_SOURCE_FILES
+        Common.cpp
+        Constants.cpp
+        VmaReplay.cpp
+        VmaUsage.cpp
+    )
+
+    add_executable(VmaReplay ${VMA_REPLAY_SOURCE_FILES})
```
adam-sawicki-a, line 11: VmaSample executable has /MP option used while this one doesn't. I think we should configure both the same way.

adam-sawicki-a:
VMA_DYNAMIC_ and STATIC_VULKAN_FUNCTIONS are not mutually exclusive. They can be both enabled at the same time, although not very useful. But it should be possible to enable none of them, as there is a 3rd option - to pass pointers to Vulkan functions via `VmaAllocatorCreateInfo::pVulkanFunctions`.

I will be out of office until July 9th, then I can merge the pull request.

src/VmaReplay/CMakeLists.txt
IAmNotHanni: Done.

IAmNotHanni:
Because `VMA_DYNAMIC_VULKAN_FUNCTIONS` is already defined in VmaUsage.hpp:

```cpp
//#define VMA_HEAVY_ASSERT(expr) assert(expr)
//#define VMA_USE_STL_CONTAINERS 1
//#define VMA_DEDICATED_ALLOCATION 0
//#define VMA_DEBUG_MARGIN 16
//#define VMA_DEBUG_DETECT_CORRUPTION 1
//#define VMA_DEBUG_INITIALIZE_ALLOCATIONS 1
//#define VMA_RECORDING_ENABLED 1
//#define VMA_DEBUG_MIN_BUFFER_IMAGE_GRANULARITY 256
//#define VMA_USE_STL_SHARED_MUTEX 0
//#define VMA_DEBUG_GLOBAL_MUTEX 1
//#define VMA_MEMORY_BUDGET 0
#define VMA_STATIC_VULKAN_FUNCTIONS 0
#define VMA_DYNAMIC_VULKAN_FUNCTIONS 1
```

I will remove it as CMake option to prevent macro redefinition.

IAmNotHanni:
Well another option would be:

```cpp
#ifndef VMA_STATIC_VULKAN_FUNCTIONS 
#define VMA_STATIC_VULKAN_FUNCTIONS  0 // This will only be set if no CMake is used
#endif
```

So which option would you prefer?

adam-sawicki-a:
The macros that we will control from Cmake can be completely removed from VmaUsage.hpp.

IAmNotHanni:
I added `VMA_STATIC_VULKAN_FUNCTIONS` and set it to `OFF` by default.
Also I changed `VMA_DYNAMIC_VULKAN_FUNCTIONS` to be `ON` by default. Is that correct?

adam-sawicki-a:
Yes, that is the way it works currently.

adam-sawicki-a:
Thank you for this major contribution! I merged it and I will make some small fixes. If you see something worth changing in the future, please feel free to open new issues or pull requests.

Issue #175, opened 2021-06-01
Check if certain VMA macro settings break build


IAmNotHanni:
When defining `VMA_USE_STL_CONTAINERS`, I get this on gcc 9.3.0:

[gcc-errors-use-STL-containers.log](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/files/6580960/gcc-errors-use-STL-containers.log)

There could be more issues with other macros. If I enable `VMA_RECORDING_ENABLED`, I get a lot of warnings as well.

IAmNotHanni:
Enabling `VMA_USE_STL_CONTAINERS` breaks the build on Windows and Linux, because STL's [std::vector](https://en.cppreference.com/w/cpp/container/vector) has no second argument `freeMemory`:

```cpp
void resize(size_t newCount, bool freeMemory = false)
{
    if(newCount > N && m_Count > N)
    {
        // Any direction, staying in m_DynamicArray
        m_DynamicArray.resize(newCount, freeMemory); // Doesn't work with std::vector
    }
    else if(newCount > N && m_Count <= N)
    {
        // Growing, moving from m_StaticArray to m_DynamicArray
        m_DynamicArray.resize(newCount, freeMemory); // Doesn't work with std::vector
        if(m_Count > 0)
        {
            memcpy(m_DynamicArray.data(), m_StaticArray, m_Count * sizeof(T));
        }
    }
    else if(newCount <= N && m_Count > N)
    {
        // Shrinking, moving from m_DynamicArray to m_StaticArray
        if(newCount > 0)
        {
            memcpy(m_StaticArray, m_DynamicArray.data(), newCount * sizeof(T));
        }
        m_DynamicArray.resize(0, freeMemory); // Doesn't work with std::vector
    }
    else
    {
        // Any direction, staying in m_StaticArray - nothing to do here
    }
    m_Count = newCount;
}
```

Here's a logfile with gcc-9.3.0: [gcc-STL-vector-errors.log](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/files/6584977/gcc-STL-vector-errors.log)

Here's a logfile from clang 10.0.0-4ubuntu1: [clang-STL-vector-errors.log](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/files/6585007/clang-STL-vector-errors.log)

PR #176, opened 2021-06-02
Fix VMA_USE_STL_CONTAINERS macro option

IAmNotHanni:
See https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/175
This fixes build for gcc-9.3.0, clang 10.0.0-4ubuntu1, and Visual Studio 2019 with `VMA_USE_STL_CONTAINERS` enabled.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -5311,13 +5311,27 @@ public:
     {
         if(newCount > N && m_Count > N)
         {
+#if VMA_USE_STL_CONTAINERS
+            m_DynamicArray.resize(newCount);
+            if(freeMemory) {
+                m_DynamicArray.shrink_to_fit();
+            }
+#else
             // Any direction, staying in m_DynamicArray
             m_DynamicArray.resize(newCount, freeMemory);
+#endif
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for reporting this issue. I prefer to add `shrink_to_fit` to `VmaVector` to make the interface matching STL.

IAmNotHanni:
Making `VmaVector` match the STL interface would not fix the build though, because if `VMA_USE_STL_CONTAINERS` is enabled, the STL is used and its `.resize()` method does not have a 2nd argument.

```cpp
// This does not work if m_DynamicArray is a std::vector, as there is no 2nd argument accepted by STL.
m_DynamicArray.resize(newCount, freeMemory);
```

adam-sawicki-a:
Did I leave `VmaVector::resize` supporting or called with 2nd argument anywhere in the code? I can't find it. Also compilation with `VMA_USE_STL_CONTAINERS=1` works for me in Visual Studio 2019. Please let me know what specific problems you can find with the latest code version.

PR #177, opened 2021-06-08
Fix clang -Wdeprecated-copy.

pkasting:
The copy constructor must be explicitly defined (here, defaulted) to be
available when the assignment operator is deleted.  While there is a
templated constructor from VmaStlAllocator<U>, the compiler doesn't
consider this to be a "copy" constructor for purposes of the warning.


adam-sawicki-a:

PR #183, opened 2021-07-13
Include block IDs in debug log messages

PazerOP:
Some existing `VMA_DEBUG_LOG` statements did not specifically mention the IDs of the blocks to which they were referring. This PR adds block IDs to the associated debug log messages.


adam-sawicki-a:
Thank you for this contribution. I implemented it the way you suggested.

wolfgangfengel:
@adam-sawicki-a this is for The Forge. We can also give you access to GitLab to check this out.
@PazerOP is working on this from our side.

PR #186, opened 2021-07-23
Add include_directories to CMake to include the header files.

HindrikStegenga:
This change makes sure the right vulkan headers can be include using `#include<vulkan/vulkan.h>` and makes sure that the imported headers are those from FindPackage() rather than those found in system include paths. (Currently VMA does not build on macOS for example as the default header paths do not include vulkan headers, even when sdk is installed correctly.)


adam-sawicki-a:

PR #187, opened 2021-07-25
Do not fail travis build if the build directory already exists.

phil-el:
Linux build failed because the build directory already exists,. Now the build will be really done so it's possible there is some other cause of failure for a travis linux build.


adam-sawicki-a:

PR #188, opened 2021-07-27
Remove unused variable.

phil-el:
Allow to increase the warning level by removing all unused variable for a default build, I didn't try to play with the various VMA_DEBUG_*. I didn't change the warning level in the build machinery. This change is dedicated to project wanting to use a more usual warning level in their own project like -Wall -pedantic -Werror

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -9563,7 +9563,6 @@ void VmaBlockMetadata_Generic::PrintDetailedMap(class VmaJsonWriter& json) const
         m_Suballocations.size() - (size_t)m_FreeCount, // allocationCount
         m_FreeCount); // unusedRangeCount
 
-    size_t i = 0;
     for(const auto& suballoc : m_Suballocations)
     {
         if(suballoc.type == VMA_SUBALLOCATION_TYPE_FREE)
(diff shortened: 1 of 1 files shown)


include/vk_mem_alloc.h
```diff
@@ -17570,8 +17569,8 @@ struct AtomicTransactionalIncrement
 VkResult VmaAllocator_T::AllocateVulkanMemory(const VkMemoryAllocateInfo* pAllocateInfo, VkDeviceMemory* pMemory)
 {
     AtomicTransactionalIncrement<uint32_t> deviceMemoryCountIncrement;
-    const uint64_t prevDeviceMemoryCount = deviceMemoryCountIncrement.Increment(&m_DeviceMemoryCount);
 #if VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT
+    const uint64_t prevDeviceMemoryCount = deviceMemoryCountIncrement.Increment(&m_DeviceMemoryCount);
```
turol, line 17573: This is wrong, it will now skip the increment if the debug option is not on.

adam-sawicki-a: Cases like this are exactly why declare I am not going to be free of any compiler warnings. Also not going to make the code more complicated (like capturing the returned value to a variable only #if...). I prefer to just silence such warnings.

turol: You could add a `VMA_DEBUG_ASSERTED` macro that expands to `__attribute__((unused))` when debugging is disabled so the compiler will shut up but if you remove the debug use will then complain.

adam-sawicki-a: This is what I mean by complicating the code just to silence compiler warnings. I'm sorry but this is explicitly defined as out-of-scope of this library in the documentation.

PR #195, opened 2021-09-22
Remove trailing whitespace

akien-mga:
Just a minor change to prevent unwanted diffs when vendoring this lib in other codebases (due to `.editorconfig` or IDE configuration).

---

BTW, this repo uses a mix of files with LF (Unix-like) line endings and CRLF (DOS-like). I would suggest standardizing on LF like most repos do (which can still be checked out as CRLF on Windows and converted to LF on push). If this is wanted, I can work on a PR to do that.


adam-sawicki-a:
Thank you for this suggestion. I standardized line endings to LF.

PR #197, opened 2021-09-29
Use forwarding references with std::forward()

danielkrupinski:
As advised in C++ Core Guidelines:
[**F.19: For “forward” parameters, pass by TP&& and only std::forward the parameter**](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f19-for-forward-parameters-pass-by-tp-and-only-stdforward-the-parameter)


adam-sawicki-a:
Thank you for this contribution and the explanation.

PR #207, opened 2021-11-16
Compatibility fix with old sdks where VK_ERROR_UNKNOWN is not defined

scribam:
`VK_ERROR_UNKNOWN` was added in the first Vulkan 1.2 SDK (https://github.com/KhronosGroup/Vulkan-Headers/releases/tag/v1.2.131). The define is to maintain the compatibility with the previous sdks.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -150,6 +150,10 @@ available through VmaAllocatorCreateInfo::pRecordSettings.
     #include <vulkan/vulkan.h>
 #endif
 
+#if !defined(VK_VERSION_1_2)
+    #define VK_ERROR_UNKNOWN ((VkResult)-13)
+#endif
+
 // Define this macro to declare maximum supported Vulkan version in format AAABBBCCC,
 // where AAA = major, BBB = minor, CCC = patch.
 // If you want to use version > 1.0, it still needs to be enabled via VmaAllocatorCreateInfo::vulkanApiVersion.


adam-sawicki-a:
Documentation says `VK_ERROR_UNKNOWN` is part of Vulkan 1.0. If this is not correct, is there a ticket for the bug in their documentation?
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap3.html#fundamentals-returncodes

scribam:
Documentation is right: in the latest SDK, `VK_ERROR_UNKNOWN` is part of Vulkan 1.0.
But, if you use a SDK older than 1.2.131 (e.g. v1.1.130), `VK_ERROR_UNKNOWN` is not available in vulkan_core.h

https://github.com/KhronosGroup/Vulkan-Headers/blob/v1.2.131/include/vulkan/vulkan_core.h#L135-L137
```
    VK_ERROR_FRAGMENTED_POOL = -12,
    VK_ERROR_UNKNOWN = -13,
    VK_ERROR_OUT_OF_POOL_MEMORY = -1000069000,
```
https://github.com/KhronosGroup/Vulkan-Headers/blob/v1.1.130/include/vulkan/vulkan_core.h#L135-L136
```
    VK_ERROR_FRAGMENTED_POOL = -12,
    VK_ERROR_OUT_OF_POOL_MEMORY = -1000069000,
```

adam-sawicki-a:
OK, in this case I think you are right with the proposed change. Thank you for this contribution.

PR #208, opened 2021-11-22
Fixes for compile errors due to unused variables

JustSid:
This simple PR just removes a couple unused variables and adds a macro that can be used to eat potentially unused variables. For example  when they are only used in an assertion. This solves compile errors in code bases that have unused variables as an error. Since VMA is supposed to be a header only library, it's impossible to not inherit these kinds of compiler error flags from the parent project (aside from using awkward, per compiler, pragmas). I figured it'd make sense to keep the VMA codebase clean in this case.

I'm not sure what the general preference here is, but it might make sense to enable this sort of error in the unit test project in order to not re-introduce it. If you want, I can add another commit to this PR to enable that.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -279,6 +279,11 @@ available through VmaAllocatorCreateInfo::pRecordSettings.
     #endif
 #endif
 
+// Macro that will evaluate anything to nothing.
+// The purpose is to work around variables that are potentially unused,
+// for example when used in an assertion that can be compiled out.
+#define VMA_POTENTIALLY_UNUSED(x) (void)(x)
+
 /** \struct VmaAllocator
 \brief Represents main object of this library initialized.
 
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
I am sorry but I will not complicate the code just to silence some compiler warnings in some project configurations on some compilers. This has been proposed many times already. Documentation clearly says in the chapter "Features not supported":
> Code free of any compiler warnings. Maintaining the library to compile and work correctly on so many different platforms is hard enough. Being free of any warnings, on any version of any compiler, is simply not feasible. There are many preprocessor macros that make some variables unused, function parameters unreferenced, or conditional expressions constant in some configurations. The code of this library should not be bigger or more complicated just to silence these warnings. It is recommended to disable such warnings instead.

adam-sawicki-a:
Are there any warnings that appear when including VMA header with declarations only, without including implementation?

PR #209, opened 2021-11-22
Added X-Plane to the list of software using VMA

JustSid:
The X-Plane Vulkan renderer is build around VMA as its memory allocation backend and has been shipping since late 2019. We make heavy use of the defragmentation feature and have very aggressive streaming of world data as well as texture paging. Planes are quite fast and earth is quite large, so we constantly move memory around and dynamically adjust things as the memory budget grows or shrinks and the user moves around. It's absolutely rock solid on top of VMA :)


adam-sawicki-a:

PR #214, opened 2021-12-07
fix crash in VmaAllocator_T::AllocateDedicatedMemory

yaoyao-cn:
move variable dedicatedAllocInfo outside the if block in case of destruct of dedicatedAllocInfo
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/master/include/vk_mem_alloc.h#L15498


adam-sawicki-a:
Thank you for finding and fixing it so quickly!

PR #224, opened 2022-01-24
Fix build - 64 bit functions do not exist

TheLavaBlock:
Hi, see topic again 🤠


adam-sawicki-a:

By the way, I am curious what is your use case that you compile 32-bit code on Windows?

TheLavaBlock:
Thanks - You're probably right 👍 

It's no longer necessary, but as long as it's still available and not completely out of the world, it can still be supported, right?

adam-sawicki-a:
Yes :)

PR #227, opened 2022-01-30
Proposal: allow custom includes without having to edit vk_mem_alloc.h

laurelkeys:
Currently, if someone whishes to redefine the macros in the `CONFIGURATION SECTION` they may need to edit "vk_mem_alloc.h" or to be careful to always include a custom header file before it. This might even make some of the currently included headers unnecessary, as it is already noted:
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/08759cbec52557f46f53f77bfd1e6bf59188c643/include/vk_mem_alloc.h#L2535-L2541

By adding a new `VMA_CONFIGURATION_USER_INCLUDES_H` macro, which defines the path to a user-configured header file, users are no longer required to make direct changes to the file (e.g. for removing the standard headers listed above).

This then simplifies the process of updating "vk_mem_alloc.h" locally.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -2533,12 +2533,33 @@ VmaAllocatorCreateInfo::pVulkanFunctions. Other members can be null.
 #endif
 
 /*
-Following headers are used in this CONFIGURATION section only, so feel free to
+Define this macro to include custom header files without having to edit this file directly, e.g.:
+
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for this contribution. I think this is a good idea, especially as this new macro can be used only in the CPP file which pulls `VMA_IMPLEMENTATION` rather than before every `#include` of the VMA header.

PR #233, opened 2022-02-07
Include vulkan.h before #define VMA_VULKAN_VERSION

laurelkeys:
Hi, upon updating my local fork my compilation started hitting the following assert:

https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/4d65f7e113480be158f85e6f924b3d775e215a2b/include/vk_mem_alloc.h#L14582-L14587

This happens because in commit https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/commit/5c710e86a0ce0664bbc2374c39b956b9928b2f29 the definition of `VMA_VULKAN_VERSION` was moved before the include of `<vulkan/vulkan.h>`. So, instead of `1002000` it is now being set to `1000000` (on the `#else` branch), since I'm not including vulkan.h myself in the `.cpp` file where I define `VMA_IMPLEMENTATION`.

https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/4d65f7e113480be158f85e6f924b3d775e215a2b/include/vk_mem_alloc.h#L132-L142

I have opened this PR because I didn't see a particular reason for the order switch in commit https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/commit/5c710e86a0ce0664bbc2374c39b956b9928b2f29, but please let me know if this is something I should correct on my usage of the library (e.g. by defining `VMA_VULKAN_VERSION` myself).

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -126,6 +126,10 @@ for user-defined purpose without allocating any real GPU memory.
 extern "C" {
 #endif
 
+#ifndef VULKAN_H_
+    #include <vulkan/vulkan.h>
+#endif
+
 // Define this macro to declare maximum supported Vulkan version in format AAABBBCCC,
 // where AAA = major, BBB = minor, CCC = patch.
 // If you want to use version > 1.0, it still needs to be enabled via VmaAllocatorCreateInfo::vulkanApiVersion.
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
I moved the definition of `VMA_VULKAN_VERSION` above functions like `extern PFN_vkGetInstanceProcAddr vkGetInstanceProcAddr;` but I forgot about this include. Thanks for fixing this.

laurelkeys:
Hey @adam-sawicki-a thank you! I was wondering if including `vulkan.h` itself is necessary, or if `vk_mem_alloc.h` could simply do:
```cpp
#ifndef VK_PLATFORM_H_
    #include <vulkan/vk_platform.h>
#endif
#ifndef VULKAN_CORE_H_
    #include <vulkan/vulkan_core.h>
#endif
```

I have tested this locally on Windows but haven't digged into the code to see if there are any dependencies on other specific platforms (that are included by `VK_USE_PLATFORM_*` in `vulkan.h`). If there aren't any though, I think it could be a good replacement, since this would avoid including `windows.h` when `VK_USE_PLATFORM_WIN32_KHR` is defined.

PR #240, opened 2022-02-18
Fix VmaSmallVector::push_back

matusfedorko:
`push_back ` on an empty `VmaSmallVector` skipped the first element


adam-sawicki-a:

PR #246, opened 2022-02-25
Add Godot Engine to the list of known usages

RandomShaper:


adam-sawicki-a:
Thank you for this pull request. I was just about to introduce the same change :)

PR #251, opened 2022-03-25
Change: Use std::popcount for VMA_COUNT_BITS_SET in C++20

spnda:
`VMA_COUNT_BITS_SET` is not optimized anymore by default and I don't think the user should themselves provide a optimization manually, especially when this isn't noted in the changelog (or in the docs?) and the stdlib provides a simple to use implementation.

I think it's appropriate to automatically use [`std::popcount`](https://en.cppreference.com/w/cpp/numeric/popcount) if C++20 is available, as it will include its own optimized implementation from the compiler, which will probably be more optimized than `VmaCountBitsSet`. Additionally, if the project is built with AVX or SSE4 flags enabled, the stdlib automatically uses `__popcnt()` or `__builtin_popcount()` (or at least I observed this with MSVC).
And I think any built-in optimization is welcome.

Also, should this change be reflected in the documentation for `VmaCountBitsSet`?
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/42be483bd5c6605e789e011aac684e0b95d05359/include/vk_mem_alloc.h#L3166-L3176

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -2766,8 +2766,13 @@ static void vma_aligned_free(void* VMA_NULLABLE ptr)
 #endif
 
 #ifndef VMA_COUNT_BITS_SET
-    // Returns number of bits set to 1 in (v)
-    #define VMA_COUNT_BITS_SET(v) VmaCountBitsSet(v)
+    #if __cplusplus >= 202002L // C++20
+        #include <bit>
+        #define VMA_COUNT_BITS_SET(v) std::popcount(v);
+    #else
+        // Returns number of bits set to 1 in (v)
+        #define VMA_COUNT_BITS_SET(v) VmaCountBitsSet(v)
+    #endif
 #endif
 
 #ifndef VMA_BITSCAN_LSB


adam-sawicki-a:
Thank you for this suggestion. I added usage of `std::popcount`, although differently than you proposed in the pull request.

spnda:
Next time a review or comment before closing would've been nice, so that a PR submitter can adapt and fix any issues, and that stuff like this doesn't look like one is stealing code. Would've thought that most larger projects, especially something coming from AMD, would share the same FOSS concepts. In the end I was the one with the idea and initial implementation, but as this is such a small change I'm not gonna be too annoyed, however I'm just concerned in general that this might happen with other contributions.

Apart from that, I actually personally prefer my solution because it doesn't waste a function call in debug, doesn't unnecessarily include the `<bit>` header if another custom implementation was already defined, and better shows that it's using `std::popcount` instead of the custom implementation. But I guess you do you, you don't seem to answer any other PRs after they got closed or merged anyway.

adam-sawicki-a:
I'm sorry, you are right, I could have left the ticket open. It was closed automatically as I typed "Closes #251" in the commit message. Please note the commit mentions this ticket number, so anyone looking at the commit history can find out it was your idea. I didn't intent to look like I stole it.

spnda:
Thank you for acknowledging this and sorry if I might've sounded a bit aggressive, was just a bit confused and surprised. I don't quite care as this change is minimal. Although I've also thought that in these cases a "Co-authored-by" in the commit metadata could be a solution, but I'm not asking for anything and I'll just leave this as it is now.

PR #254, opened 2022-03-27
C4062 fix

MennoVink:
I'm compiling vma from inside my own project which uses warnings to errors and has C4062 enabled. This switch statement wont compile because it thinks the Cleanup operation isn't being handled.


MennoVink:
Added another commit that is needed to keep xcode happy. It complained about possible misuse of the comma operator at ln 5884

adam-sawicki-a:

PR #255, opened 2022-03-30
Added vmaGetAllocatorAllocationCallbacks

NongusStudios:
Allows the user to get the VkAllocationCallbacks.


adam-sawicki-a:
I am sorry but I don't want to add more of `vmaGet*` type of functions to the library API. Please note that Vulkan itself doesn't provide this kind of functions either - you can only set the current state of things, not get it back. The user of the library who passes certain parameters should remember them somewhere if they are needed at a later time.

NongusStudios:
Fair enough. I just wanted this because in ImGui, in the initialisation info it asks for allocation callbacks in the allocator field and there was no other way of getting them without adding this function.

PR #259, opened 2022-04-19
Fix return type of GetAllocationNextPtr

brianosman:
Returning `void* const` doesn't mean anything (and produces a warning with -Wignored-qualifiers). Based on usage, it looks like the returned pointer is expected to point at const data, so I've fixed the declaration to reflect that.


adam-sawicki-a:
You are right. Thank you for this contribution.

PR #264, opened 2022-05-13
Expose notPreferredFlags

Neo-Zhixing:
Can we possibly expose `notPreferredFlags` in `VmaAllocationCreateInfo` so that this could be specified by the application?

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -1238,6 +1238,11 @@ typedef struct VmaAllocationCreateInfo
     Set to 0 if no additional flags are preferred. \n
     If `pool` is not null, this member is ignored. */
     VkMemoryPropertyFlags preferredFlags;
+    /** \brief Flags that preferably should not be set in a memory type chosen for an allocation.
+
+    Set to 0 if no flags are undesired. \n
+    If `pool` is not null, this member is ignored. */
+    VkMemoryPropertyFlags notPreferredFlags;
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
I am sorry, but this has been considered before as not-to-be-done. If you want more advanced way of selecting memory type, please choose eligible memory types manually and set `VmaAllocationCreateInfo::memoryTypeBits`.

Neo-Zhixing:
@adam-sawicki-a I understand that it was previously decided in #110 not to do this, but with the release of VMA 3.0 our selection logic is already implementing `nonPreferredFlag`, and `VMA_MEMORY_USAGE_AUTO`, VMA_MEMORY_USAGE_AUTO_PREFER_DEVICE` and `VMA_MEMORY_USAGE_AUTO_PREFER_HOST` are also already using `nonPreferredFlag`. This change does not complicate the logic any more than they already are - it merely expose this flag to the end user so that they can do the same in their own selection logic.

I'm trying to build an alternative selection algorithm by accepting an enum of common memory allocation scenarios in my library. Using `memoryTypeBits` here will significantly complicate the logic.

I would appreciate it if we could reconsider the decision we made 2 years ago. Thanks!

adam-sawicki-a:
You could do it in two steps:
- Try to allocate with `memoryTypeBits` set to only the memory types that don't contain the non-preferred flag.
- If that failed, try to allocate with `memoryTypeBits` set to all the eligible memory types.

PR #272, opened 2022-06-14
Convert VMA line endings to Unix style (LF)

akien-mga:
This makes it consistent with other files in the repository,
and is typically the norm for cross-platform Git repos.

Just ran `dos2unix` on the file, no code change.


adam-sawicki-a:

PR #273, opened 2022-06-14
Strip trailing whitespace in code files

akien-mga:
Will conflict with #272 for the VMA header, I'll rebase as needed.


akien-mga:
> Will conflict with #272 for the VMA header, I'll rebase as needed.

Rebased.

adam-sawicki-a:
Thank you for this contribution. I merged it.

However, I don't consider it an important issue. Due to the way the library is developed I can't guarantee that trailing spaces or Windows line endings won't appear in the future.

PR #274, opened 2022-06-14
[cmake] Updated installation process to use modern CMake targets

pnarvor:
This is a slight modification of the installation process using modern CMake targets.

Modifications follow CMake recommendations : https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html

This allows to easily integrate VulkanMemoryAllocator in other CMake-based projects (looking as follow)

`find_package(VulkanMemoryAllocator REQUIRED)
target_link_libraries(yourtarget PUBLIC VulkanMemoryAllocator::VulkanMemoryAllocator)
`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -24,7 +24,9 @@ set_target_properties(
     CXX_STANDARD_REQUIRED ON
 )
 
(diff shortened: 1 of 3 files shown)


adam-sawicki-a:

PR #275, opened 2022-06-19
Fix windows.h not found when cross compiling on Linux

Attractadore:
The header provided by mingw-w64 on Linux is called windows.h.


adam-sawicki-a:

PR #277, opened 2022-07-25
docs: Fixed typo under 'Selecting Vulkan version' doc section

Max-Rodriguez:
Just a small grammar correction! Found it while I was reading the quick start guide :)


adam-sawicki-a:
HTML documentation is generated automatically by Doxygen tool from special comments inside file vk_mem_alloc.h. We should not modify those HTML files as they will be overwritten anyway next time the documentation is generated. It should be fixed in vk_mem_alloc.h.

Max-Rodriguez:
> HTML documentation is generated automatically by Doxygen tool from special comments inside file vk_mem_alloc.h. We should not modify those HTML files as they will be overwritten anyway next time the documentation is generated. It should be fixed in vk_mem_alloc.h.

Oh gotcha 👍 Will correct the commit in a couple hours

Max-Rodriguez:
I've updated the branch to make the typo fix at the header file

adam-sawicki-a:

PR #282, opened 2022-08-25
Add vcpkg installation instructions

JonLiu1993:
VulkanMemoryAllocator is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for VulkanMemoryAllocator and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build VulkanMemoryAllocator, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/VulkanMemoryAllocator/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)


JonLiu1993:
@adam-sawicki-a, Thanks for pointing out the mistakes and making corrections. If you want to add other libraries, you can refer to this [document ](https://github.com/microsoft/vcpkg/blob/master/docs/examples/packaging-github-repos.md)and add them in vcpkg yourself, or you can submit an issue in vcpkg for other contributors to add them

PR #285, opened 2022-09-07
build: Revert addition of aligned_alloc()

brad0:
VulkanMemoryAllocator builds as C++14. aligned_alloc() was only added in C++17. C headers are not guarenteed to expose the function until C++17 mode, as noted on OpenBSD, but other implementations such as glibc do the same.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -2693,7 +2693,9 @@ remove them if not needed.
    #define VMA_NULL   nullptr
 #endif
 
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
You are right - `aligned_alloc`  was only added in C++17. I pushed a change that uses the function only in C++17, asserts and returns null if no implementation of `vma_aligned_alloc` can be determined.

However, I don't understand your proposed pull request. Why do you propose to delete some code, e.g. related to Apple platform? Some other users added it to fix their platforms. If I accept your pull request, it will likely break for them.

PR #287, opened 2022-09-12
allow building against vulkan 1.2.131 headers

paroj:
e.g. on ubuntu 20.04


adam-sawicki-a:

PR #304, opened 2022-11-29
Minor CMake cleanup

juan-lunarg:
Broken up into 3 atomic commits

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ message(STATUS "VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT = ${VMA_DEBUG_
 
 if(VMA_BUILD_SAMPLE)
     set(VMA_BUILD_SAMPLE_SHADERS ON)
-endif(VMA_BUILD_SAMPLE)
+endif()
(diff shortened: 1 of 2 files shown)


adam-sawicki-a:
What is the minimum Cmake version supporting `MSVC_IDE`?

adam-sawicki-a:
Cmake documentation says:
> Set to true when the target platform is the Microsoft Visual C++ IDE, as opposed to the command line compiler.

Is this intended that the settings you propose to put unde `MSVC_IDE` would work only with this variable set? What if the user wants to use MSVC command line compiler? Then, I guess they wouldn't be applied?

juan-lunarg:
> What is the minimum CMake version supporting `MSVC_IDE`?

It's supported by 3.9, this is an old variable.

juan-lunarg:
> Is this intended that the settings you propose to put unde `MSVC_IDE` would work only with this variable set? What if the user wants to use MSVC command line compiler? Then, I guess they wouldn't be applied?

My intent was to match the existing code. I've updated the code so it's friendlier to Ninja using MSVC.

adam-sawicki-a:
Thank you very much for this contribution!

PR #310, opened 2023-01-04
Fix atomic type used in AtomicTransactionalIncrement

orlowd:
This effectively reverts some of the changes from commit 4dfa169ffc65583cd35bdb7ec13057cde8897c66, pertaining to the `AtomicTransactionalIncrement` class. The problem with them is that they broke compilation when a custom atomic type is defined with `VMA_ATOMIC_UINT32`.

In `VmaAllocator_T::AllocateVulkanMemory()` a call to `AtomicTransactionalIncrement::Increment()` is given as argument a reference to the `m_DeviceMemoryCount`. But the latter will be of the user-defined `VMA_ATOMIC_UINT32` type, while the method expects to be passed a value of the `std::atomic` type.

Reverting those changes has fixed the compilation error with custom atomic types defined.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -4006,6 +4006,8 @@ template<typename AtomicT>
 struct AtomicTransactionalIncrement
 {
 public:
+    using T = decltype(AtomicT().load());
+
     ~AtomicTransactionalIncrement()
     {
         if(m_Atomic)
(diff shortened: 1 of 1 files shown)


orlowd:
Oops, it seems that this change broke CI.

Considering [this stack overflow question](https://stackoverflow.com/questions/53648614/what-happened-to-stdatomicxvalue-type) it looks like older compilers that are used with CI do not implement a defect report that retrofits the value_type member field into std::atomic. So I’ve added a separate commit that uses a workaround mentioned there that seems to fix it.

adam-sawicki-a:

PR #320, opened 2023-02-02
Fix compilation warning.

jeroenbakker-atmind:
Detected during inclusion of this library in Blender when compiling using GPU 11.2.0/Linux.
The compiler suggests to add the parentheses around the '&&', but makes more sense
to add them around the '||' as the message is to inform the developer about the issue.


adam-sawicki-a:
Thank you for this contribution. Indeed it was a bug.

PR #321, opened 2023-02-14
Add GitHub actions to address GNU/Clang warnings

juan-lunarg:
The intent here is to address warnings coming from `include/vk_mem_alloc.h` and make sure it stays fixed by having it as part of CI.

NOTE: By default VMA_BUILD_ERROR is OFF to not cause package managers any trouble.

Split up into atomic commits to highlight fixing the compiler warnings.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -23,7 +23,6 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
         -Wno-unused-parameter
         -Wno-unused-variable
         -Wno-missing-field-initializers
-        -Wno-implicit-fallthrough
     )
 endif()
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,6 +8,25 @@ add_library(VulkanMemoryAllocator
     ${PROJECT_SOURCE_DIR}/include/vk_mem_alloc.h
 )
 
+option(VMA_BUILD_WERROR "Treat compiler warnings as errors")
+if (VMA_BUILD_WERROR)
+    target_compile_options(VulkanMemoryAllocator PRIVATE
+        "$<IF:$<CXX_COMPILER_ID:MSVC>,/WX,-Werror>"
+    )
+endif()
+
+if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
+    target_compile_options(VulkanMemoryAllocator PRIVATE
+        -Wpedantic
+        -Wall
+        -Wextra
+        -Wno-unused-parameter
+        -Wno-unused-variable
+        -Wno-missing-field-initializers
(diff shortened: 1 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -8,6 +8,24 @@ add_library(VulkanMemoryAllocator
...
+endif()
+
+if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
+    target_compile_options(VulkanMemoryAllocator PRIVATE
+        -Wpedantic
+        -Wall
+        -Wextra
+        -Wno-unused-parameter
+        -Wno-unused-variable
+        -Wno-missing-field-initializers
```
juan-lunarg, line 25: My goal is to fix the remaining warnings in another PR. I don't want to complicate this one too much.

juan-lunarg:
I'd like to add some rational to this.

While some users consume VMA via system include commands like `-isystem dir` not all do. And many users don't even know about `-isystem dir`. So they just silence warnings in their project after introducing a dependency.

And for MVSC the problem is even worse because until very recently they didn't even have an equivalent to `-isystem`.

So the best thing for project to do is to make sure that their public interface is as clean as possible. Which is what I'm trying to do here.

adam-sawicki-a:
Thank you for this proposal.

Successful build of the library is already validated in Windows + Visual Studio:
https://ci.appveyor.com/project/adam-sawicki-amd/vulkanmemoryallocator/branch/master
As well as Linux + GCC and Clang:
https://app.travis-ci.com/github/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator

We may consider enabling GitHub actions instead in the future.

Regarding compiler warnings, documentation explicitly says in chapter "Features not supported":

> Code free of any compiler warnings. Maintaining the library to compile and work correctly on so many different platforms is hard enough. Being free of any warnings, on any version of any compiler, is simply not feasible. There are many preprocessor macros that make some variables unused, function parameters unreferenced, or conditional expressions constant in some configurations. The code of this library should not be bigger or more complicated just to silence these warnings. It is recommended to disable such warnings instead.

The change you proposed that adds `break;` in two places actually changes code logic and makes it incorrect. Fallthrough is intentional there.

turol:
> The change you proposed that adds `break;` in two places actually changes code logic and makes it incorrect. Fallthrough is intentional there.

You should add a "fallthrough" comment to those. It both tells the reader that it's supposed to fall through and silences the warning.

turol:
Nope, GCC and Clang (I think) will interpret a comment containing the word "fallthrough" the same and shut up about it.

adam-sawicki-a:
I added macro `VMA_FALLTHROUGH` non-empty only when C++17 is enabled + a comment with word "Fallthrough". Please let me know if it works for you.

juan-lunarg:
That handles it nicely :+1:

PR #325, opened 2023-03-22
Change __alignof to alignof

lawadr:
Whereas __alignof is a non-standard compiler extension that seems to come from Microsoft and is supported (yet undocumented) by Clang and GCC, alignof is a standard language feature in C++11. Clang and GCC both document __alignof__ as a compiler extension, but this isn't supported by Microsoft.

This fixes builds on compilers that don't support __alignof.


adam-sawicki-a:

PR #326, opened 2023-04-02
IWYU: add cstdio for snprintf in vk_mem_alloc.h

stha09:
Discoverd with upcoming GCC 13.


adam-sawicki-a:
`#include <cstdio> // For snprintf` was already there under `#if VMA_STATS_STRING_ENABLED` and so are all uses of `snprintf` in "vk_mem_alloc.h". Why do you propose to include this file unconditionally?

stha09:
In chromium there is an old copy in chromium tree. I will try to trigger an update there. Feel free to revert.

PR #329, opened 2023-04-11
Add missing cstdio include

Mystro256:
I noticed this issue when compiling on GCC 13.0.1. GCC complains because "snprintf" is defined by cstdio, but it's not included.


adam-sawicki-a:
Likely duplicate of #326. Please use latest version of the code from "master" branch, not the latest Release, which hasn't been updated for long time.

Mystro256:
@adam-sawicki-a thanks, I was using master, but looks like I didn't pull it correctly to get this fix locally. I'll close this.

PR #340, opened 2023-05-24
Extended VmaAllocationInfo to include the size of the actual allocation

debaetsd:
Extended VmaAllocationInfo to include the actual allocation size (as-in the size used to allocate the memory when calling VkAllocateMemory).

We need the size when performing interop between Vulkan and OpenGL.
`glImportMemoryFdEXT` requires the size of the actual underlying Vulkan object to be passed in (not the size of a suballocated resource).

We worked around it by using dedicated allocations though extending VMA to query the size of the underlying allocation allows the benefits of the block allocator.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -1365,6 +1365,13 @@ typedef struct VmaAllocationInfo
     `vkCmdCopyBuffer` must be limited to the size of the resource.
     */
     VkDeviceSize size;
+    /** \brief Size of this actual allocation (as made to vkAllocateMemory), in bytes.
+
+    It never changes.
+
+    \note Allocation size returned in this variable is always larger or equal to VmaAllocationInfo::size.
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for reporting this issue. I am sorry but I don't want to merge it. Query for `VmaAllocationInfo` is a function that may be called very often, so I want to keep it fast and avoid growing it in scope, while returning the size of the underlying memory block looks to me like an exotic feature.

Please feel free to maintain your fork of the code with this change. Another option could be using dedicated allocations or custom pools with specific `VmaPoolCreateInfo::blockSize` so that the size is known.

I also suggest to rename the new parameter, as "allocation" is used in VMA to name a single sub-allocated region of a memory block, as in `VmaAllocation`, so I recommend to rather call it a "block", "memory block", or "Vulkan memory" size.

PR #341, opened 2023-05-31
cmake: correctly use find_dependency

abouvier:
`find_package(VulkanMemoryAllocator)` (without `REQUIRED`) should not fail if `Vulkan` is not found.


adam-sawicki-a:
How does this change interact with newly merged #345?

abouvier:
I updated this PR for the new logic.

Issue #303, opened 2022-11-28
Do not import Vulkan package from CMakeLists.txt


tomilov:
Use [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) cmake option or [this](https://github.com/KhronosGroup/Vulkan-Headers/blob/main/CMakeLists.txt#L32) workaround to determine whether it fine to call `find_package(Vulkan)`. If VMA is submodule (`add_subdirectory`) of some project as well as Vulkan-Headers, then it is poison to inject some system-wide Vulkan headers from top level `CMakeLists.txt`.

adam-sawicki-a:
Please note VMA is intended to be used as a single-header STB-style library with file "vk_mem_alloc.h" just copied to your project source code. By including this entire repository as a submodule and compiling VMA as a separate library you clutter your project with unnecessary files, add unnecessary complexity, and limit the ability to reconfigure VMA by changing its source code or predefining some `VMA_` macros.

tomilov:
In case of strait file copying user loose ability to use `git submodule` command functionality to keep `vk_mem_alloc.h` up to date with origin repo. I mean there are use cases when namely git submodule way to add VMA to a project. These are important part of all use cases. In case of import VMA via `add_subdirectory` VMA project `CMakeLists.txt` can export single target (say named `vma` or `VulkanMemoryAllocator`, as repo named), having just [INTERFACE_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html) property set. To inject `Vulkan-Headers` unconditionally is not on the right way for the case.

adam-sawicki-a:
I added usage of `PROJECT_IS_TOP_LEVEL` as you recommended. Please validate if it works for you.

tomilov:
Thank you. I think this should work.

tomilov:
`include_directories(${Vulkan_INCLUDE_DIR})` have to be also under the guards

tomilov:
`find_package(Doxygen)` can be after `if(BUILD_DOCUMENTATION)` before `if(DOXYGEN_FOUND)`

tomilov:
In the case checking `PROJECT_IS_TOP_LEVEL` for `Doxygen` is not necessarily required

adam-sawicki-a:
[...]

PR #345, opened 2023-06-28
cmake: always add vulkan headers dir

abouvier:
Vulkan headers directory is always needed because `vulkan/vulkan.h` is included in `src/VmaUsage.h` and `include/vk_mem_alloc.h`.
And the fix for https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/303 was wrong. If `Vulkan::Headers` is already available when VulkanMemoryAllocator is added as a subdirectory, then `find_package(Vulkan)` in `VulkanMemoryAllocator/CMakeLists.txt` will not be executed.
Example code:
```
add_subdirectory(Vulkan-Headers)
set(VMA_STATIC_VULKAN_FUNCTIONS OFF)
set(VMA_DYNAMIC_VULKAN_FUNCTIONS ON)
add_subdirectory(VulkanMemoryAllocator) # VulkanMemoryAllocator will use the headers from the above directory
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,14 +2,6 @@ cmake_minimum_required(VERSION 3.9)
 
 project(VulkanMemoryAllocator LANGUAGES CXX)
 
-# https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
-string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_TOP_LEVEL)
-
-if(PROJECT_IS_TOP_LEVEL)
-    find_package(Vulkan REQUIRED)
-    include_directories(${Vulkan_INCLUDE_DIR})
-endif()
-
 # VulkanMemoryAllocator contains an sample application which is not built by default
 option(VMA_BUILD_SAMPLE "Build VulkanMemoryAllocator sample application" OFF)
 option(VMA_BUILD_SAMPLE_SHADERS "Build VulkanMemoryAllocator sample application's shaders" OFF)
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -26,9 +26,11 @@ target_include_directories(VulkanMemoryAllocator PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
 )
 
-# Only link to Vulkan if static linking is used
+# Only link to Vulkan library if static linking is used, but always add Vulkan headers directory
 if(${VMA_STATIC_VULKAN_FUNCTIONS})
     target_link_libraries(VulkanMemoryAllocator PUBLIC Vulkan::Vulkan)
+else()
+    target_include_directories(VulkanMemoryAllocator PUBLIC "${Vulkan_INCLUDE_DIRS}")
```
Tachi107, line 33: Shouldn't this be `target_link_libraries(VulkanMemoryAllocator PUBLIC Vulkan::Headers)`?

adam-sawicki-a:
I am fine with increasing the minimum required version of Cmake if needed.

CMakeLists.txt
```diff
@@ -31,6 +23,12 @@ message(STATUS "VMA_DEBUG_INITIALIZE_ALLOCATIONS = ${VMA_DEBUG_INITIALIZE_ALLOCA
 message(STATUS "VMA_DEBUG_GLOBAL_MUTEX = ${VMA_DEBUG_GLOBAL_MUTEX}")
 message(STATUS "VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT = ${VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT}")
 
+if(VMA_STATIC_VULKAN_FUNCTIONS OR NOT TARGET Vulkan::Headers)
+    find_package(Vulkan REQUIRED)
+else()
+    get_target_property(Vulkan_INCLUDE_DIRS Vulkan::Headers INTERFACE_INCLUDE_DIRECTORIES)
+endif()
+
```
juan-lunarg, line 31: Seems simpler to move this code to `src/CMakeLists.txt`

abouvier: Why? It's better if a missing required dependency halts the build sooner. And imported targets found in the top CMakeLists.txt make them available in subdirectories if necessary.

abouvier: You can't merge both codes, the condition is not exactly the same in each file :p
The linking logic is different from the finding logic.
This is what the correct merge would look like (note the double `find_package`):
```
if(VMA_STATIC_VULKAN_FUNCTIONS)
    find_package(Vulkan REQUIRED)
    target_link_libraries(VulkanMemoryAllocator PUBLIC Vulkan::Vulkan)
else()
    if (TARGET Vulkan::Headers)
        get_target_property(Vulkan_INCLUDE_DIRS Vulkan::Headers INTERFACE_INCLUDE_DIRECTORIES)
    else()
        find_package(Vulkan REQUIRED)
    endif()
    target_include_directories(VulkanMemoryAllocator PUBLIC "${Vulkan_INCLUDE_DIRS}")
endif()
```
Also if you need to link with the Vulkan library in multiple subdirectories you will do this from the top `CMakeLists.txt`:
```
find_package(Vulkan REQUIRED)
add_subdirectory(src)
add_subdirectory(test)
...
```
And not execute `find_package` in each subdirectory.

abouvier:
I changed the logic without having to raise the minimal version.
I you want to ignore the `find_package` and use your own Vulkan library/headers located in a subdirectory, you can use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) with `OVERRIDE_FIND_PACKAGE`.

PR #347, opened 2023-06-30
Add attribute to mark pNext pointers with the Vulkan struct they extend

Rob2309:
This PR adds an attribute `VMA_EXTENDS_VK_STRUCT` similar to `VMA_LEN_IF_NOT_NULL` that allows detecting programatically which Vulkan structure will be extended with a given `pNext` argument.

For example:
```c++
VMA_CALL_PRE VkResult VMA_CALL_POST vmaBindBufferMemory2(
    VmaAllocator VMA_NOT_NULL allocator,
    VmaAllocation VMA_NOT_NULL allocation,
    VkDeviceSize allocationLocalOffset,
    VkBuffer VMA_NOT_NULL_NON_DISPATCHABLE buffer,
    const void* VMA_NULLABLE VMA_EXTENDS_VK_STRUCT(VkBindBufferMemoryInfoKHR) pNext);
```
with this attribute, generators can detect and enforce that `pNext` points to a structure that is allowed to be present in the `pNext` chain of `VkBindBufferMemoryInfoKHR`.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -234,6 +234,10 @@ extern "C" {
     #define VMA_CALL_POST
 #endif
 
+#ifndef VMA_EXTENDS_VK_STRUCT
+    #define VMA_EXTENDS_VK_STRUCT(vkStruct)
+#endif
+
 // Define this macro to decorate pointers with an attribute specifying the
 // length of the array they point to if they are not null.
 //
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
How is it going to be used? Can you please show an example of how it is used and why it is needed?

Rob2309:
> How is it going to be used? Can you please show an example of how it is used and why it is needed?

I would take the `ash` rust bindings as an example.
In ash, there is a trait called [`ExtendsBindBufferMemoryInfo`](https://docs.rs/ash/latest/ash/vk/trait.ExtendsBindBufferMemoryInfo.html) that is only implemented for structs that are allowed to be included in the `pNext` chain of `VkBindBufferMemoryInfo` as per the Vulkan specification.

With the attribute introduced above, a code generator could automatically generate a typesafe wrapper for `vmaBindBufferMemory2` that only allows passing an allowed structure as `pNext`.

adam-sawicki-a:
OK, thank you for the explanation. I accepted and merged your change without any modifications.

Issue #308, opened 2022-12-24
Vma c interface issue


bensam00:
I have included vk_mem_alloc.h on my c project and in my cmake file i added it as an interface like this 

`add_library(include INTERFACE)`
`target_include_directories(include INTERFACE include)`

while include is the folder where i put my vma header file 
and i added it without `#define VMA_IMPLEMENTATION` as mention in documentation like this 
` #include "include/vk_mem_alloc.h"`

and yet it always gives me 
`undefined reference to 'vmaCreateAllocator'`
when i try to initialize it 

im using fedora 37 with vulkan 1.3

adam-sawicki-a:
"Undefined reference" likely means that you didn't include VMA implementation in your project. Cmake script is provided as an external contribution. Recommended way to use the library is to add "vk_mem_alloc.h" file directly to your project. Please follow instructions from documentation https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/quick_start.html. Long story short, in exactly one C++ source file of your code you need to #include the VMA header preceded by `#define VMA_IMPLEMENTATION` so the internal implementation of the library is compiled. Only then your project will include implementation of functions declared by the header.

mr-sihc:
`mold: error: /usr/lib64/libVulkanMemoryAllocator.a(VmaUsage.cpp.o):(.text): R_X86_64_PC32 relocation at offset 0x1987 against symbol 'vtable for VmaBlockMetadata' can not be used; recompile with -fPIC`
i get this error, i'm already compiling and linking with -fPIC
but its probably the library that didnt compile and link with -fPIC
can you add this to the CMake?

adam-sawicki-a:
This library is developed mostly on and for Windows + Visual Studio. I don't know what is `-fPIC` so I am not sure whether I should add this option to CMake.

Please try to include "vk_mem_alloc.h" directly in your project code, like the documentation recommends, instead of using it as a separate library with CMake and see if it helps on your platform.

mr-sihc:
i did -fPIC mean position independent code
which libraries usually get compiled with

juan-lunarg:
I believe this issue can now be closed with all the CMake fixes made to master.

-fPIC doesn't apply at all anymore now that `GPUOpen::VulkanMemoryAllocator` is a minimal header only library like `Vulkan::Headers`.

[...]

PR #349, opened 2023-07-05
CMake Overhaul

### LINKED ISSUE #346: CMake installation issue: Most users just want `vk_mem_alloc.h`
I think this issue is worth bringing up since I've seen the same pattern in 2 places. When shipping the Vulkan SDK and with vcpgk.

NOTE: I believe Conan also doesn't use the default installation either (https://github.com/conan-io/conan-center-index/tree/master/recipes/vulkan-memory-allocator/all). But I'm not as familiar with Conan.

### Problem: Most users just want `vk_mem_alloc.h`

Here is the current port code for vcpkg:
```
vcpkg_from_github(
    OUT_SOURCE_PATH SOURCE_PATH
    REPO GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator
    REF a6bfc237255a6bac1513f7c1ebde6d8aed6b5191 #v3.0.1
    SHA512  14361ff201fd660c22b60de54c648ff20a2e2a7f65105f66853a9a4dbffbeca2ae42098dcb1528bb4e524639b92fa4ff27ebd3940c42ccfaf7c99c08bdd0d8ce
    HEAD_REF master
    PATCHES
        gcc13.patch # https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/commit/29d492b60c84ca784ea0943efc7d2e6e0f3bdaac
)

file(COPY "${SOURCE_PATH}/include/vk_mem_alloc.h" DESTINATION "${CURRENT_PACKAGES_DIR}/include")

file(COPY "${CMAKE_CURRENT_LIST_DIR}/unofficial-vulkan-memory-allocator-config.cmake" DESTINATION "${CURRENT_PACKAGES_DIR}/share/unofficial-vulkan-memory-allocator")

configure_file("${SOURCE_PATH}/LICENSE.txt" "${CURRENT_PACKAGES_DIR}/share/${PORT}/copyright" COPYONLY)
```

It's ignoring the installation method natively provided by `VulkanMemoryAllocator` and providing it's own method instead.

Same applies to the Vulkan SDK. We also ignore the installation provided by `VulkanMemoryAllocator` and just manually copy the include file. We don't want to ship binaries for the `VulkanMemoryAllocator`.

juan-lunarg:
Dramatically simplify the CMake code for end users.

- Tested samples build correctly on Windows
- Only installs vk_mem_alloc.h and a minimal CMake config file
- CMake 3.23 to utilize FILE_SET and simplify sample code
- Removes lots of CMake code reducing maintenance issues
- Uses `Vulkan::` namespace so it's similar to VulkanHeaders `Vulkan::Headers`
- Very friendly to add_subdirectory/find_package users
- Easier to use since you don't have to worry about all the CMake configuration options affecting the binary

closes #346


CMakeLists.txt
```diff
@@ -1,53 +1,47 @@
...
-message(STATUS "VMA_DEBUG_INITIALIZE_ALLOCATIONS = ${VMA_DEBUG_INITIALIZE_ALLOCATIONS}")
-message(STATUS "VMA_DEBUG_GLOBAL_MUTEX = ${VMA_DEBUG_GLOBAL_MUTEX}")
-message(STATUS "VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT = ${VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT}")
-
-if(VMA_STATIC_VULKAN_FUNCTIONS OR VMA_BUILD_SAMPLE)
-    find_package(Vulkan REQUIRED)
-else()
-    find_package(VulkanHeaders REQUIRED)
-endif()
+cmake_minimum_required(VERSION 3.23...3.26)
```
juan-lunarg, line 1: Note: If requested I can lower the CMake version required. FILE_SET was introduced in `3.23` which makes the CMake code a lot easier to read/write for a header only library.

CMakeLists.txt
```diff
@@ -1,53 +1,47 @@
...
-        COMMENT "Generating API documentation with Doxygen"
-        VERBATIM )
+option(VMA_INSTALL "Install Vulkan Memory Allocator" ${PROJECT_IS_TOP_LEVEL})
+if (VMA_INSTALL)
+    include(GNUInstallDirs)
+
+    set_target_properties(VulkanMemoryAllocator PROPERTIES EXPORT_NAME "MemoryAllocator")
+
+    install(TARGETS VulkanMemoryAllocator EXPORT VulkanMemoryAllocator FILE_SET vk_mem_alloc)
+    install(EXPORT VulkanMemoryAllocator NAMESPACE "Vulkan::" DESTINATION "${CMAKE_INSTALL_DATADIR}/cmake/VulkanMemoryAllocator")
```
juan-lunarg, line 22: I tested that the following find_package support worked without issue. I can add testing to CI if requested.

```
find_package(VulkanMemoryAllocator CONFIG)
target_link_libraries(foobar PRIVATE Vulkan::MemoryAllocator)
```

CMakeLists.txt
juan-lunarg, line 1: My overall goal is to make this project easier to install and match expectations of various package managers so they can remove their workarounds.

Also removing as much obstacles as possible that makes this code difficult to maintain.

README.md
```diff
@@ -97,41 +97,6 @@ With this one function call:
...
-| VmaSample | VMA sample application | `VMA_BUILD_SAMPLE` | `OFF` |
-| VmaBuildSampleShaders | Shaders for VmaSample | `VMA_BUILD_SAMPLE_SHADERS` | `OFF` |
-
-Please note that while VulkanMemoryAllocator library is supported on other platforms besides Windows, VmaSample is not.
-
-These CMake options are available
-
-| CMake option | Description | Default setting |
-| ------------- | ------------- | ------------- |
-| `VMA_RECORDING_ENABLED` | Enable VMA memory recording for debugging | `OFF` |
```
juan-lunarg, line 126: VMA_RECORDING_ENABLED is not a CMake option. In general my advice is to NOT list CMake options in documentation since they tend to age.

src/CMakeLists.txt
```diff
@@ -1,107 +1,76 @@
...
 )
 
-if (MSVC)
-    # Provides MSVC users nicer debugging support
-    target_sources(VulkanMemoryAllocator PRIVATE ${CMAKE_CURRENT_LIST_DIR}/vk_mem_alloc.natvis)
-endif()
-
-set_target_properties(
-    VulkanMemoryAllocator PROPERTIES
+target_link_libraries(VmaUsage PUBLIC Vulkan::MemoryAllocator)
```
juan-lunarg, line 30: Note. This is similar to `Vulkan::Headers` which is the intent.

juan-lunarg (commented):
Please let me know what you think I'm receptive to anything. I just want to make your job easier and write the simplest CMake possible to not.

CMakeLists.txt
```diff
@@ -1,53 +1,47 @@
...
+            COMMAND ${DOXYGEN_EXECUTABLE} ${DOXYGEN_OUT}
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+            COMMENT "Generating API documentation with Doxygen"
+            VERBATIM
+        )
+    endif()
+
+    option(VMA_BUILD_SAMPLES "Build samples")
+    if (VMA_BUILD_SAMPLES)
+        add_subdirectory(src)
```
juan-lunarg, line 45: It might be pragmatic to change the directory name to `samples` instead of `src`.

src/CMakeLists.txt
```diff
@@ -1,107 +1,76 @@
...
+message(STATUS "VMA_DEBUG_INITIALIZE_ALLOCATIONS = ${VMA_DEBUG_INITIALIZE_ALLOCATIONS}")
+message(STATUS "VMA_DEBUG_GLOBAL_MUTEX = ${VMA_DEBUG_GLOBAL_MUTEX}")
+message(STATUS "VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT = ${VMA_DEBUG_DONT_EXCEED_MAX_MEMORY_ALLOCATION_COUNT}")
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+add_library(VmaUsage STATIC)
+target_sources(VmaUsage PRIVATE
```
juan-lunarg, line 25: Highlight a potential usage of vk_mem_alloc.h as a stb-style header instead of trying to enforce it on the user. Also since it's not being installed anymore just make it STATIC due to DLLs on Windows being a royal pain due to lack of RPATH support.

juan-lunarg:
Should also close https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/308

Since VMA will become a header only dependency that doesn't have to think about position independent code compiler options. That responsibility will fall to the user instead.

CMakeLists.txt
juan-lunarg: Here is what the VulkanHeaders CMake code looks like for reference (Which has 3.15 as the minimum):
https://github.com/KhronosGroup/Vulkan-Headers/blob/main/CMakeLists.txt

juan-lunarg: It's worth pointing out though that the find_package support is backwards compatible:

image

So vcpkg/conan/etc users won't be affected by the CMake minimum required.

CMakeLists.txt
```diff
@@ -1,53 +1,44 @@
...
-if(VMA_STATIC_VULKAN_FUNCTIONS OR VMA_BUILD_SAMPLE)
-    find_package(Vulkan REQUIRED)
-else()
-    find_package(VulkanHeaders REQUIRED)
+cmake_minimum_required(VERSION 3.23...3.26)
+
+project(VMA LANGUAGES CXX)
+
+add_library(VulkanMemoryAllocator INTERFACE)
+add_library(Vulkan::MemoryAllocator ALIAS VulkanMemoryAllocator)
```
juan-lunarg, line 6: Note. It may be better to use `gpuopen::` instead of `vulkan::` since this repo technically isn't inside the KhronosGroup.

I don't have a preference. But it would simplify the lines of code on 18/21.

abouvier:
Aren't you supposed to link the interface library `VulkanMemoryAllocator` with `Vulkan::Headers`? Because `vk_mem_alloc.h` includes `vulkan/vulkan.h`.

juan-lunarg:
Closing since it seems like CMake support will be removed in the future.

abouvier:
> Furthermore vk_mem_alloc.h has the following lines of code:
> 
> ```
> #ifndef VULKAN_H_
>     #include <vulkan/vulkan.h>
> #endif
> ```
> 
> Meaning if vulkan.h was already included by the parent project it won't include it itself.

I don't see the point of this condition, because it's already (as in all headers) the first line of `vulkan/vulkan.h`. So if `vulkan/vulkan.h` has already been included, then this new include is a no-op.

> Yet we don't declare a public dependency on the X11 libraries.

Well you should, because instead of doing:

```
find_package(VulkanMemoryAllocator)
target_link_libraries(MyApp PRIVATE GPUOpen::VulkanMemoryAllocator)
```

One need to do:

```
find_package(VulkanMemoryAllocator)
find_package(VulkanHeaders)
find_package(X11)
...
target_link_libraries(Vulkan::Headers INTERFACE X11::Xrandr)
target_link_libraries(GPUOpen::VulkanMemoryAllocator INTERFACE Vulkan::Headers)
...
target_link_libraries(MyApp PRIVATE GPUOpen::VulkanMemoryAllocator)
```

Or, if `GPUOpen::VulkanMemoryAllocator` is used only once:

```
find_package(VulkanMemoryAllocator)
find_package(VulkanHeaders)
find_package(X11)
...
target_link_libraries(MyApp PRIVATE GPUOpen::VulkanMemoryAllocator Vulkan::Headers X11::Xrandr ...)
```

If the location of these headers is not added to the include directories of `MyApp` then the build will fail.

It works easy right now for most users because headers are almost every time installed in standard directories and so are found without this interdependence otherwise needed.

juan-lunarg:
That's more than fine. Considering there are multiple ways to get/consume Vulkan-Headers, X11, XCB, Wayland, etc.

Not everyone uses `find_package(VulkanHeaders)` and not everyone uses `find_package(X11)`.

It's impossible to make everyone happy. Everyone has different use cases. Being permissive is necessary.

PR #350, opened 2023-07-06
CMake Overhaul

### LINKED ISSUE #346: CMake installation issue: Most users just want `vk_mem_alloc.h`

juan-lunarg:
Dramatically simplify the CMake code for end users.

- Tested samples build correctly on Windows
- Only installs vk_mem_alloc.h and a minimal CMake config file
- CMake 3.23 to utilize FILE_SET and simplify sample code
- Removes lots of CMake code reducing maintenance issues
- Uses `Vulkan::` namespace so it's similar to VulkanHeaders `Vulkan::Headers`
- Very friendly to add_subdirectory/find_package users
- Easier to use since you don't have to worry about all the CMake configuration options affecting the binary

closes #346


juan-lunarg:
Comments from this PR still apply:
https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/349

I can't re-open a closed PR.

adam-sawicki-a:
Thank you for this contribution. I merged your pull request. However, I had some issues with it, so I made few fixes. Please let me know what you think of them.

- Removed separate library "VmaUsage", merged these things into "VmaSample". Or should we leave it? Is it used externally?
- Added the main file "vk_mem_alloc.h" to the list of source files so it appears in Visual Studio solution. It was missing.
- Set `VS_DEBUGGER_WORKING_DIRECTORY` so that the app launches correctly from Visual Studio debugger after building. Without it, it couldn't find compiled shader files, as starting directory was "bin/src" instead of "bin".
- Removed redundant `#ifndef VULKAN_H_` around Vulkan include, as you advised.
- Changed project name from "VMA" back to "VulkanMemoryAllocator".

adam-sawicki-a:
OK, thank you for the quick feedback. Any further fixes and improvements from you are welcome!

Regarding project name, I agree it is not ideal to have project name == library name, but this is how it worked before, so changing it breaks CI build here, possibly others' builds as well. "VMA" is also a bit short and maybe could conflict with some other projects.

juan-lunarg:
> In particular GPUOpen::VulkanMemoryAllocator provides a very unique name that is not likely to cause any conflicts.

Namespaces in CMake provide a very similar affect to namespace in C++. Preventing collisions for users.

adam-sawicki-a:
I see what you mean. OK, I changed it to "VMA" as you proposed.

PR #351, opened 2023-07-10
fixed copy paste error in vulkan sample

lobneroO:


adam-sawicki-a:

PR #352, opened 2023-07-11
cmake: Use share instead of DATADIR for CMake config files

juan-lunarg:
See the following discussion for context:
https://discourse.cmake.org/t/what-should-the-destination-be-for-a-header-only-librarys-cmake-config-file/8473/2

NOTE: I've never seen this cause an issue before, but better safe than sorry. Also I think it reads a bit better.

I recently did this for VulkanHeaders so I'd like this fixed here as well.


juan-lunarg:
> I recently did this for VulkanHeaders so I'd like this fixed here as well.

https://github.com/KhronosGroup/Vulkan-Headers/pull/422

adam-sawicki-a:

PR #353, opened 2023-07-11
Fixup include of "vk_mem_alloc.h"

juan-lunarg:
No need to include via `../include`

By linking against `GPUOpen::VulkanMemoryAllocator` the proper include directory is added to the project.


adam-sawicki-a:

PR #356, opened 2023-07-19
cmake: install docs

abouvier:
The documentation is generated but not installed.
And if only the project name was still `VulkanMemoryAllocator` then
```
install(DIRECTORY docs/ DESTINATION "${CMAKE_INSTALL_DATADIR}/doc/VulkanMemoryAllocator" PATTERN ".nojekyll" EXCLUDE)
```
could have been simply:
```
install(DIRECTORY docs/ TYPE DOC PATTERN ".nojekyll" EXCLUDE)
```
By the way,
```
install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
can be simplified to:
```
install(DIRECTORY include/ TYPE INCLUDE)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ if (PROJECT_IS_TOP_LEVEL)
             COMMENT "Generating API documentation with Doxygen"
             VERBATIM
         )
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
@juan-lunarg What do you think about the proposed changes?

juan-lunarg:
> @juan-lunarg What do you think about the proposed changes?

LGTM. Gives package managers the option to install the documentation if they want to.

adam-sawicki-a:

Issue #357, opened 2023-07-23
Overly agressive deallocation of free memory blocks


Attractadore:
Say we have 2 sets of frame resources that don't fit into one block of memory, but fit into two.
At the beginning of every frame, all the resources from 2 frames back are deleted and reallocated again.
In my use case, this leads to a severe slowdown, since every odd frame VMA calls vkFreeMemory and than vkAllocateMemory.

I suspect that in my case the memory allocation would happen some thing like this:
1. Allocate resources for even frame in block 0
2. Allocate resources for odd frame in block 0 and block 1
3. Free resources for previous even frame in block 0 and allocate them again from block 0
4. Free resources for previous odd frame in blocks 0 and 1, free block 1, allocate some resources from block 0, allocate block 1 again, allocate some resources from block 1

I tried disabling lines 12695 through 12705 (in VmaBlockVector::Free), and that fixed things for me, since VMA stopped constantly calling vkFreeMemory and vkAllocateMemory.
I think that the heuristic to delete a block that is already empty whenever an allocation is freed is not very good, since depending on the order in which allocations are freed, it can easily lead to blocks constantly being allocated and deallocated.
So maybe it should be disabled.

Another solution that I thought of was adding an option that would allow memory pools to never free empty blocks until explicitly asked to do so. But that would require more time to design and implement.

Attractadore:
Created PR #358 to disable the heuristic

adam-sawicki-a:
Creating and destroying images or buffers every frame is not a recommended way of using Vulkan. Games and other graphics applications tend to create such resources at startup, when loading a level, worst case in runtime when a new one is needed (preferably asynchronous to render frame, in the background, on a separate thread) and then use them for rendering every frame.

adam-sawicki-a:
If you want to ensure certain number of memory blocks to stay allocated when you free your resources, you can also use a custom pool with non-zero `VmaPoolCreateInfo::minBlockCount`.

PR #358, opened 2023-07-23
Disable deallocation of already empty blocks on when freeing an allocation 

Attractadore:


adam-sawicki-a:
I am sorry but I don't think this is a good idea. Saving GPU memory is the highest priority. Please see my explanation in #357.

PR #364, opened 2023-09-01
Make C++20 features optional

manas-kulkarni:
- Some compilers with C++20 dont have std::popcount
- Support to toggle C++20 features from user side using new VMA_CPP20 macro. If the macro is not defined, it will be set using default logic - #if __cplusplus >= 202002L || _MSVC_LANG >= 202002L


adam-sawicki-a:

PR #396, opened 2024-01-12
Fix sub-allocation check in the wraparound path of the linear allocator

dm-tesla:
Was this check correct in the first place? Nothing in this section of code actually relies (in a hard way) on the suballocations being empty. In the case I hit, the `suballocations1st` was allocated (for the capacity of 8) but emtpy, while `suballocations2nd` wasn't allocated.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -9088,7 +9088,7 @@ bool VmaBlockMetadata_Linear::CreateAllocationRequest_LowerAddress(
     // beginning of 1st vector as the end of free space.
     if (m_2ndVectorMode == SECOND_VECTOR_EMPTY || m_2ndVectorMode == SECOND_VECTOR_RING_BUFFER)
     {
-        VMA_ASSERT(!suballocations1st.empty());
+        VMA_ASSERT(suballocations1st.data() != nullptr);
 
         VkDeviceSize resultBaseOffset = 0;
         if (!suballocations2nd.empty())


dm-tesla:
I don't know if checking for `data()` is the correct thing to do here. I haven't reviewed the algorithm in full and only looked at the parts locally, where I see that having a non-empty vector doesn't seem to be required.
The motivation for this change is me seeing this assertion triggering for no apparent reason (from our side). If there is a better solution to this, I'll happily accept it :)

adam-sawicki-a:
Do you have some test code or description of steps to take to reproduce this bug, this assert triggering for no reason?

dm-tesla:
It's only failing in a complex environment with many allocations happening, and not immediately either. So far, I've seen it on a 10Mb allocation and on a 80Mb allocation. I don't have a simplified test case to provide here, sorry!

The best way to reproduce it cleanly would just construct a test that would randomly allocate and free data using this allocator, using chunks big enough to reach this code path. Looking at the test infrastructure here, it's not clear to me yet if it is comprehensive enough to trigger the failure. Do you have code coverage info from the test runs?

adam-sawicki-a:
Thank you for reporting this bug. This assert is correct. Control flow should never reach this place if `suballocations1st` vector is empty. Below you can see references to its element `suballocations1st[index1st]`, which could crash is it was empty.

I was able to reproduce this assert hitting when using a virtual allocator and trying to make an allocation larger than the block. I implemented a test and a fix for it. Please let me know if it solves the problem for you.

If the problem still occurs:

- Please let me know if it happens for you on a virtual allocator or normal Vulkan memory allocation (a custom pool).
- Please ensure `VMA_HEAVY_ASSERT` also asserts for testing purposes and check if it is not hit, especially on `Validate`, which would indicate corrupted state of the object.
- Do you use the linear allocator in the ring buffer fashion, or as a stack, or double stack? You should not mix those methods.
- If not talking about the virtual allocator, but a custom pool, do you limit the number of blocks in the pool? With ring buffer or double stack mode, you should use `minBlockCount = maxBlockCount = 1`. Otherwise, the behavior is undefined.

PR #403, opened 2024-01-25
Add VK_API_VERSION definitions for compatibility with older Vulkan headers

adrianlopezroche:
Older Vulkan SDK headers lack the required definitions for `VK_API_VERSION_VARIANT`, `VK_API_VERSION_MAJOR`, `VK_API_VERSION_MINOR`, and `VK_API_VERSION_PATCH`. This causes compilation errors on some platforms, especially those whose official repositories don't contain the latest versions of the Vulkan SDK. This patch checks for these definitions and defines them if necessary. The definitions are taken directly from the SDK, version 1.3.275.0.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -3247,6 +3247,20 @@ static const uint32_t VMA_ALLOCATION_INTERNAL_STRATEGY_MIN_OFFSET = 0x10000000u;
 static const uint32_t VMA_ALLOCATION_TRY_COUNT = 32;
 static const uint32_t VMA_VENDOR_ID_AMD = 4098;
 
+// VK_API_VERSION definitions for compatibility with older Vulkan headers.
+#ifndef VK_API_VERSION_VARIANT
+#define VK_API_VERSION_VARIANT(version) ((uint32_t)(version) >> 29U)
+#endif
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Where do you see a use of the new `VK_API_VERSION_` macros in VMA code? I cannot find it anywhere in the latest version of "master" branch, so it doesn't seem to be a problem.

I would prefer to just keep using the old `VK_VERSION_` macros. Is there any benefit of using the new ones and treating the 2 most significant bits as a new "variant" part?

adrianlopezroche:
I was working from the latest release (v3.0.1) which uses the `VK_API_VERSION_*` variants and mistakenly assumed it was the same in master. I guess I figured there was some reason for using the newer definitions, so I didn't even bother to check if they were still being used. The Vulkan headers say the `VK_VERSION_*` variants are deprecated, but I don't know if that matters much.

It's probably OK to close this and leave it the way it is unless there's reason to care about the old define being deprecated.

adam-sawicki-a:
OK, thank you for the explanation. I don't see a reason to use the new macros if the old ones are good enough and compatible with old Vulkan.

Please use latest code from the "master" branch. It is always kept in a good shape. Official release was created more than a year ago. There were many bug fixes and other changes since then.

PR #404, opened 2024-01-26
Remove semicolon outside function

serboupal:
Remove this semicolon to fix ISO C warnings


adam-sawicki-a:

PR #408, opened 2024-02-12
Add macros for leak reporting.

res2k:
Allows for more customizability of the leak messages.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -3019,6 +3019,29 @@ static void vma_aligned_free(void* VMA_NULLABLE ptr)
     #define VMA_LEAK_LOG_FORMAT(format, ...)   VMA_DEBUG_LOG_FORMAT(format, __VA_ARGS__)
 #endif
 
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
`VMA_LEAK_LOG_FORMAT` macro was introduced recently to distinguish logging about memory leaks from other logging. I am sorry, but I don't want to complicate this further to avoid feature creep. "vk_mem_alloc.h" file is already very long.

res2k:
@adam-sawicki-a, does that mean you don't like the additional macros, or that you're not open to _any_ means to obtain/output leak information apart from the log messages as they are now?
(_Some_ way to customize or replace `DebugLogAllocation()`, I guess ...)

FWIW, I have seen projects that generate "single header implementations" from multiple source files.
(Unfortunately I don't recall exactly where I've seen that. But I found https://github.com/JamesBoer/Heady, ie automation for this scenario exists.)
Wouldn't make the end result smaller, but perhaps makes things more manageable during development.

PR #409, opened 2024-02-13
Attempt to document leak detection aids

res2k:
Getting rid of leaks is important.

VMA has some facilities that aid in tracking down leaks, but they don't seem to be mentioned in the docs.
Attempt to provide a section on them.


adam-sawicki-a:
Thank you very much for this contribution! You are right. These macros are worth documenting.

PR #418, opened 2024-05-31
Make Vulkan include optional

manas-kulkarni:
- ```#include <vulkan/vulkan.h>``` adds dependency on include directory being set externally
- Dont include vulkan.h if ```VULKAN_H_``` is already defined (if the Vulkan header is already included)


adam-sawicki-a:

Issue #421, opened 2024-06-18
Validation layer error in sample application


IAmNotHanni:
Hello
For current main (7942b798289f752dc23b0a79516fd8545febd718), I get this validation error on `Windows 11` with `CMake 3.26.3` and `Visual Studio 17.10.3`. The cube of the sample application is being rendered correctly though. The validation layer error is printed every frame (I am only showing one here in my logs).

## Running VMA sample application on AMD Ryzen™ 9 7950X (AMD Radeon(TM) Graphics)

```
Vulkan Memory Allocator Sample 3.1.0
Vulkan API version used: 1.3
physicalDeviceProperties:
    driverVersion: 0x8000FA
    vendorID: 0x1002 (AMD)
    deviceID: 0x164E
    deviceType: 1 (INTEGRATED_GPU)
    deviceName: AMD Radeon(TM) Graphics
    limits:
        maxMemoryAllocationCount: 4096
        bufferImageGranularity: 1 B
        nonCoherentAtomSize: 128 B
physicalDeviceVulkan11Properties:
    maxMemoryAllocationSize: 2147483648 B (2.00 GB)
physicalDeviceVulkan12Properties:
    driverID: 1 (AMD_PROPRIETARY)
    driverName: AMD proprietary driver
    driverInfo: 22.40.40 (AMD proprietary shader compiler)

MEMORY HEAPS:
heapCount=3, typeCount=8
Heap 0: 268435456 B (256.00 MB) DEVICE_LOCAL, MULTI_INSTANCE
    Type 0: DEVICE_LOCAL
    Type 4: DEVICE_LOCAL, DEVICE_COHERENT (AMD), DEVICE_UNCACHED (AMD)
Heap 1: 16437477376 B (15.31 GB)
    Type 1: HOST_VISIBLE, HOST_COHERENT
    Type 3: HOST_VISIBLE, HOST_COHERENT, HOST_CACHED
    Type 5: HOST_VISIBLE, HOST_COHERENT, DEVICE_COHERENT (AMD), DEVICE_UNCACHED (AMD)
    Type 7: HOST_VISIBLE, HOST_COHERENT, HOST_CACHED, DEVICE_COHERENT (AMD), DEVICE_UNCACHED (AMD)
Heap 2: 268435456 B (256.00 MB) DEVICE_LOCAL, MULTI_INSTANCE
[...]

PR #424, opened 2024-06-21
Fix swapchain semaphore handling

IAmNotHanni:
Closes https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/421

Please verify if this solution is technically correct.
Tested on NVIDIA RTX 3090, NVIDIA GeForce 940M, Intel Arc A770, and AMD Ryzen 9 7950X.

Have a nice weekend
Johannes

### DIFF
--- a/src/VulkanSample.cpp
+++ b/src/VulkanSample.cpp
@@ -89,8 +89,11 @@ static VkCommandBuffer g_MainCommandBuffers[COMMAND_BUFFER_COUNT];
 static VkFence g_MainCommandBufferExecutedFances[COMMAND_BUFFER_COUNT];
 VkFence g_ImmediateFence;
 static uint32_t g_NextCommandBufferIndex;
-static VkSemaphore g_hImageAvailableSemaphore;
-static VkSemaphore g_hRenderFinishedSemaphore;
+// Notice we need as many semaphores as there are swapchain images
+static std::vector<VkSemaphore> g_hImageAvailableSemaphores;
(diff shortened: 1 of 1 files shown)


src/VulkanSample.cpp
```diff
@@ -2306,7 +2319,7 @@ static void DrawFrame()
 
     // Acquire swapchain image
     uint32_t imageIndex = 0;
-    VkResult res = vkAcquireNextImageKHR(g_hDevice, g_hSwapchain, UINT64_MAX, g_hImageAvailableSemaphore, VK_NULL_HANDLE, &imageIndex);
+    VkResult res = vkAcquireNextImageKHR(g_hDevice, g_hSwapchain, UINT64_MAX, g_hImageAvailableSemaphores.at(g_SwapchainImageIndex), VK_NULL_HANDLE, &imageIndex);
```
IAmNotHanni, line 2322: Maybe a comment on this here: We can't do

```cpp
uint32_t imageIndex = 0;
VkResult res = vkAcquireNextImageKHR(g_hDevice, g_hSwapchain, UINT64_MAX, g_hImageAvailableSemaphores.at(imageIndex), VK_NULL_HANDLE, &imageIndex);
```

because `imageIndex` is written to by `vkAcquireNextImageKHR` (pass by reference), so we can't use is in the call directly.
I had to track the swapchain image index manually, as can be seen at the end of `DrawFrame`.

adam-sawicki-a:
Thank you for this fix.

PR #430, opened 2024-07-10
Small maintenance

IAmNotHanni:


adam-sawicki-a:
I am sorry but I don't want to expose heap size limit as a Cmake option. That code is useful only during development of the library and I don't think it anyone using the library would want to enable it.

The removal of that outdated code about CSV recording looks good though.

adam-sawicki-a:

Issue #423, opened 2024-06-21
Use vkSetDebugUtilsObjectNameEXT to make debugging easier


IAmNotHanni:
To make debugging of the sample application easier, we should make use of [vkSetDebugUtilsObjectNameEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkSetDebugUtilsObjectNameEXT.html), especially since it's already loaded in the current sample app, but not used yet.

PR #434, opened 2024-07-10
Use SetDebugUtilsObjectName

IAmNotHanni:
Closes https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/423
This is not used in the tests yet though.

This now is the resource list in RenderDoc:
grafik

Note: We can't assign names to the physical devices which are present, but not chosen. This is why there are two resources with default names "Physical Device".

### DIFF
--- a/src/VulkanSample.cpp
+++ b/src/VulkanSample.cpp
@@ -86,7 +86,7 @@ static std::vector<VkImageView> g_SwapchainImageViews;
 static std::vector<VkFramebuffer> g_Framebuffers;
 static VkCommandPool g_hCommandPool;
 static VkCommandBuffer g_MainCommandBuffers[COMMAND_BUFFER_COUNT];
-static VkFence g_MainCommandBufferExecutedFances[COMMAND_BUFFER_COUNT];
+static VkFence g_MainCommandBufferExecutedFences[COMMAND_BUFFER_COUNT];
 VkFence g_ImmediateFence;
 static uint32_t g_NextCommandBufferIndex;
 // Notice we need as many semaphores as there are swapchain images
@@ -2088,7 +2088,7 @@ static void InitializeApplication()
     fenceInfo.flags = VK_FENCE_CREATE_SIGNALED_BIT;
     for(size_t i = 0; i < COMMAND_BUFFER_COUNT; ++i)
     {
-        ERR_GUARD_VULKAN( vkCreateFence(g_hDevice, &fenceInfo, g_Allocs, &g_MainCommandBufferExecutedFances[i]) );
+        ERR_GUARD_VULKAN( vkCreateFence(g_hDevice, &fenceInfo, g_Allocs, &g_MainCommandBufferExecutedFences[i]) );
     }
 
     ERR_GUARD_VULKAN( vkCreateFence(g_hDevice, &fenceInfo, g_Allocs, &g_ImmediateFence) );
@@ -2226,10 +2226,10 @@ static void FinalizeApplication()
 
     for(size_t i = COMMAND_BUFFER_COUNT; i--; )
     {
-        if(g_MainCommandBufferExecutedFances[i] != VK_NULL_HANDLE)
+        if(g_MainCommandBufferExecutedFences[i] != VK_NULL_HANDLE)
         {
-            vkDestroyFence(g_hDevice, g_MainCommandBufferExecutedFances[i], g_Allocs);
-            g_MainCommandBufferExecutedFances[i] = VK_NULL_HANDLE;
+            vkDestroyFence(g_hDevice, g_MainCommandBufferExecutedFences[i], g_Allocs);
+            g_MainCommandBufferExecutedFences[i] = VK_NULL_HANDLE;
         }
     }
     if(g_MainCommandBuffers[0] != VK_NULL_HANDLE)
@@ -2290,7 +2290,7 @@ static void DrawFrame()
     // Begin main command buffer
     size_t cmdBufIndex = (g_NextCommandBufferIndex++) % COMMAND_BUFFER_COUNT;
     VkCommandBuffer hCommandBuffer = g_MainCommandBuffers[cmdBufIndex];
-    VkFence hCommandBufferExecutedFence = g_MainCommandBufferExecutedFances[cmdBufIndex];
+    VkFence hCommandBufferExecutedFence = g_MainCommandBufferExecutedFences[cmdBufIndex];
 
     ERR_GUARD_VULKAN( vkWaitForFences(g_hDevice, 1, &hCommandBufferExecutedFence, VK_TRUE, UINT64_MAX) );
     ERR_GUARD_VULKAN( vkResetFences(g_hDevice, 1, &hCommandBufferExecutedFence) );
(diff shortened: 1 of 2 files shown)


src/VulkanSample.cpp
```diff
@@ -261,16 +262,30 @@ struct CommandLineParameters
     }
 } g_CommandLineParameters;
 
-void SetDebugUtilsObjectName(VkObjectType type, uint64_t handle, const char* name)
+void SetDebugUtilsObjectName(VkObjectType type, uint64_t handle, std::string name)
 {
-    if(vkSetDebugUtilsObjectNameEXT_Func == nullptr)
+    TEST(!name.empty());
```
adam-sawicki-a, line 267: Why do you want to disallow empty names? I wouldn't add such check.

src/VulkanSample.cpp
```diff
@@ -261,16 +262,30 @@ struct CommandLineParameters
     }
 } g_CommandLineParameters;
 
-void SetDebugUtilsObjectName(VkObjectType type, uint64_t handle, const char* name)
+void SetDebugUtilsObjectName(VkObjectType type, uint64_t handle, std::string name)
```
adam-sawicki-a, line 265: Big objects like strings in C++ should be passed by `const&` not by value, because passing them by value involves additional object creation and copy, which involves dynamic memory allocation.

src/VulkanSample.cpp
```diff
@@ -261,16 +262,30 @@ struct CommandLineParameters
...
+    // this string ends up in the read-only data section of the program. However, if we for example want to
+    // name every item in a dynamic array of items (swapchain images for example), we can only use a std::string,
+    // and we must make sure the object lifetime of that string does not end before it's read as pObjectName.
+    // Therefore, we store the strings as a static unordered_set here just to be sure.
+    // Do not use a std::vector because it would continue to grow, eventually running out of memory.
+    static std::unordered_set<std::string> debug_names;
+    auto result = debug_names.insert(name);
+    const char* pObjectName = result.first->c_str();
+
+    if (vkSetDebugUtilsObjectNameEXT_Func == nullptr)
```
adam-sawicki-a, line 281: I think this check could be moved higher, so that we don't add the string to `debug_names` if the extension is not loaded.

src/VulkanSample.cpp
```diff
@@ -261,16 +262,30 @@ struct CommandLineParameters
     }
 } g_CommandLineParameters;
 
-void SetDebugUtilsObjectName(VkObjectType type, uint64_t handle, const char* name)
+void SetDebugUtilsObjectName(VkObjectType type, uint64_t handle, std::string name)
 {
-    if(vkSetDebugUtilsObjectNameEXT_Func == nullptr)
+    TEST(!name.empty());
+
+    // We must make sure the std::string we pass as name is still valid memory until the program ends.
```
adam-sawicki-a, line 269: Why do you assume this is the case? Which part of the Vulkan specification says the string we pass to `vkSetDebugUtilsObjectNameEXT` must be alive for the whole lifetime of the object? I can't see any mention of it here:
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDebugUtilsObjectNameInfoEXT.html
I think this is not true, the string doesn't need to persist in memory, and the VK_EXT_debug_utils Vulkan extensions makes an internal copy of the string.

adam-sawicki-a (changes requested):

src/VulkanSample.cpp
IAmNotHanni: I agree. I was introducing the policy that every Vulkan object has a non-empty name in my own codebase and applied this here as well. But Vulkan spec does not say this. I will remove that.

IAmNotHanni:
There is another small thing which I would like to mention here for completeness. As you can see in the resource list in [RenderDoc](https://renderdoc.org/), we can not name all the `VkPhysicalDevice` which are enumerated in `VulkanUsage::SelectPhysicalDevice`. First, we can't call `SetDebugUtilsObjectName` in `VulkanUsage::SelectPhysicalDevice`, because the `VkDevice` is not created yet:

```cpp
VkPhysicalDevice VulkanUsage::SelectPhysicalDevice(const GPUSelection& GPUSelection) const
{
    uint32_t deviceCount = 0;
    ERR_GUARD_VULKAN(vkEnumeratePhysicalDevices(g_hVulkanInstance, &deviceCount, nullptr));
    std::vector<VkPhysicalDevice> physicalDevices(deviceCount);
    if(deviceCount > 0)
    {
        ERR_GUARD_VULKAN(vkEnumeratePhysicalDevices(g_hVulkanInstance, &deviceCount, physicalDevices.data()));
    }

    for (size_t i = 0; i < deviceCount; i++) {
        std::string physicalDeviceName = "physicalDevices[" + std::to_string(i) + "]";
        SetDebugUtilsObjectName(VK_OBJECT_TYPE_BUFFER, reinterpret_cast<std::uint64_t>(physicalDevices[i]), "physicalDeviceName");
    }
    // ...
```

This results in:
grafik

So I thought: Let's store `std::vector<VkPhysicalDevice> physicalDevices;` as a global variable and name all physical devices after the `VkDevice` has been created?

```cpp
// Global variable
std::vector<VkPhysicalDevice> g_hPhysicalDevices;
VkPhysicalDevice VulkanUsage::SelectPhysicalDevice(const GPUSelection& GPUSelection) const
{
    uint32_t deviceCount = 0;
    ERR_GUARD_VULKAN(vkEnumeratePhysicalDevices(g_hVulkanInstance, &deviceCount, nullptr));
    if(deviceCount > 0)
    {
        g_hPhysicalDevices.resize(deviceCount);
        ERR_GUARD_VULKAN(vkEnumeratePhysicalDevices(g_hVulkanInstance, &deviceCount, g_hPhysicalDevices.data()));
    }
    // ...
```

```cpp
// ...
ERR_GUARD_VULKAN( vkCreateDevice(g_hPhysicalDevice, &deviceCreateInfo, g_Allocs, &g_hDevice) );
SetDebugUtilsObjectName(VK_OBJECT_TYPE_DEVICE, reinterpret_cast<std::uint64_t>(g_hDevice), "g_hDevice");
// Only now that SetDebugUtilsObjectName is loaded, we can assign a name to g_hVulkanInstance as well
SetDebugUtilsObjectName(VK_OBJECT_TYPE_INSTANCE, reinterpret_cast<std::uint64_t>(g_hVulkanInstance), "g_hVulkanInstance");

// Name all the physical devices enumerated in VulkanUsage::SelectPhysicalDevice
for (size_t i = 0; i < g_hPhysicalDevices.size(); i++) {
    std::string physicalDeviceName = "g_hPhysicalDevices[" + std::to_string(i) + "]";
    SetDebugUtilsObjectName(VK_OBJECT_TYPE_PHYSICAL_DEVICE, reinterpret_cast<std::uint64_t>(g_hPhysicalDevices[i]), physicalDeviceName);
}
// ...
```

This somehow crashes, but I don't know why:
grafik

I am not sure what the issue here is, but I thought I wanted to mention this.

src/VulkanSample.cpp
IAmNotHanni: Oh right, it does create a copy internally. I changes this.

IAmNotHanni:
I just pushed the requested changes instead of doing a force-push.

adam-sawicki-a:
Thank you very much for this contribution.

Let's leave physical devices unnamed then. I don't think this is a problem.

PR #435, opened 2024-07-13
Allow external control of whether to install

scribam:


adam-sawicki-a:

PR #445, opened 2024-08-28
Fix for vkGetMemoryWin32HandleKHR

Agrael1:
dumb error from me


adam-sawicki-a:
Thank you for this fix.

PR #446, opened 2024-08-28
Added a simple compilation test and linux CI (and fix linux build)

Mrkol:
Trivial compilation test is required because the sample is windows-specific. Linux CI is needed because pull requests must stop breaking linux builds. The tiny `m_Handle(VMA_NULL)` -> `m_Handle()` fix is needed to fix linux CI.

As promised @adam-sawicki-a . Same day delivery because my CI is still broken =)
(I will stop using master for my builds soon, I promise)


include/vk_mem_alloc.h
```diff
@@ -16675,7 +16675,7 @@ VMA_CALL_PRE VkResult VMA_CALL_POST vmaGetMemoryWin32Handle(VmaAllocator VMA_NOT
     VMA_DEBUG_GLOBAL_MUTEX_LOCK;
     return allocation->GetWin32Handle(allocator, hTargetProcess, pHandle);
 }
-#endif // VMA_EXTERNAL_MEMORY_WIN32 
+#endif // VMA_EXTERNAL_MEMORY_WIN32
```
adam-sawicki-a, line 16678: It seems that you change many lines that are unrelated to your change due to using some automated formatting tools. Changes like this make it difficult to analyze the history of a file later (like "blame" function).

compilation_test/vma_impl.cpp
adam-sawicki-a, line 1: Would it be possible to reuse VulkanSample app for testing compilation on Linux instead of creating a new project? VulkanSample.cpp and VmaUsage.* files already have appropriate `#ifdef`-s and even some Clang-specific `#pragma` directives.

adam-sawicki-a (changes requested):

compilation_test/vma_impl.cpp
IAmNotHanni: Does this mean we have to port VulkanSample app to Linux? I would be happy to help.

IAmNotHanni:
Does this pull request mean you want to move away from appveyor and use just GitHub actions instead? I could help with that. In my opinion, it would be nice to have a CI for Windows and Linux and both Debug and Release.

adam-sawicki-a:
@IAmNotHanni Help is very welcome. It would be good to have CI builds set up for both Windows and Linux using some free online service. We had AppVeyor for Windows and TravisCI for Linux. Unfortunately, TravisCI stopped working >2 years ago. AppVeyor is still testing Windows.

It is enough to test if the library compiles successfully (both public interface and the code under `VMA_IMPLEMENTATION`).

Running some actual tests in CI could be considered in the future, but that would require access to some Vulkan implementation on the server (a GPU or software emulation) and likely major changes in the sample app and testing code, which I would prefer to avoid. The problem is I use Windows at work and I don't have a local Linux development environment to be able to test and fix things.

include/vk_mem_alloc.h
IAmNotHanni: @Mrkol Let me know if you need help doing this :)

adam-sawicki-a:
The proposed code fails with error:

>Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

IAmNotHanni:
> The proposed code fails with error:
> 
> > Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

Use `actions/checkout@v4` like in here: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/master/.github/workflows/build.yml

adam-sawicki-a:
File ".github/workflows/linux.yaml" added by this change uses "actions/checkout@v4". I think it's the external repository "humbletim/install-vulkan-sdk@v1.1.1" referenced by this change that triggers this error. I think we should just download and install Vulkan SDK (or just its headers) without referencing such external repository.

IAmNotHanni:
> File ".github/workflows/linux.yaml" added by this change uses "actions/checkout@v4". I think it's the external repository "humbletim/install-vulkan-sdk@v1.1.1" referenced by this change that triggers this error. I think we should just download and install Vulkan SDK (or just its headers) without referencing such external repository.

Yes, it should be downloaded directly from Khronos like this for example:

```
- name: Install Vulkan SDK
  shell: bash
  run: |
    curl -LS -o vulkansdk.tar.xz https://sdk.lunarg.com/sdk/download/${{ env.VMA_VULKAN_VERSION }}/linux/vulkansdk-linux-x86_64-${{ env.VMA_VULKAN_VERSION }}.tar.xz
    echo "${{ env.VMA_VULKAN_SDK_CHECKSUM_LINUX }} vulkansdk.tar.xz" | sha256sum --check
    mkdir -p ${{ env.VMA_VULKAN_SDK_PATH }}
    tar xf vulkansdk.tar.xz -C "${{ env.VMA_VULKAN_SDK_PATH }}"
    rm -rf vulkansdk.tar.xz
```

IAmNotHanni:
@adam-sawicki-a Looking at the recent changes you pushed to main branch, one thing which points out to me is that clang and gcc seem to display no warnings at all in the CI build. In my experience, gcc and clang are much more detailed with warnings and they usually catch things which MSVC does not report by default. Maybe flags like `-Wall` and `-Wextra` are required to display any warnings?

adam-sawicki-a:
Yes, this sounds like a good idea as long as those warnings don't fail the build. Can you please propose such change?

IAmNotHanni:
Also, don't forget that in VmaSample code, there are a lot of `#indef _WIN32`, so on Linux, no code is realyl compiled from the tests. I guess the library itself is compiled though? I'm still working on https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/466 so we can have unit tests in the console on Linux :)

adam-sawicki-a:
I added Linux CI build as #477, based on this ticket. Thank you very much for your help with it.

IAmNotHanni:
> Yes, this sounds like a good idea as long as those warnings don't fail the build. Can you please propose such change?

Yes, let me see..

PR #447, opened 2024-08-29
Fix for ABI interop (without VK_KHR_external_memory_win32)

Agrael1:


GamesTrap:
I've just hit this problem in my CI/CD pipeline on Linux with GCC 13 & Clang 17.

Compiler logs: https://github.com/GamesTrap/TRAP/actions/runs/10655546793?pr=393

Agrael1:
Yep, waiting for almighty creator :)

Agrael1:
I also have the API that's waiting linux fix
https://github.com/Agrael1/Wisdom/actions/runs/10638479781

adam-sawicki-a:
Thank you for this fix.

PR #454, opened 2024-11-29
Fix ineffective error code checks.

crud89:
I believe that the error code checking in `vmaCreateImage` is wrong, possibly causing null pointer dereferencing issues, if allocation fails. The checks in this method test the result value against `>= 0`, which should always evaluate to `true`, as `VK_SUCCESS` is defined as `0` and all errors are integers greater than that. This function is the only function that tests results this way.

This PR replaces the checks by explicitly testing against `VK_SUCCESS`, which I believe is the intended behavior.

For context, I've encountered this issue, as clang-tidy complained about a possible null pointer deref here: 

https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/05973d8aeb1a4d12f59aadfb86d20decadba82d1/include/vk_mem_alloc.h#L16161-L16188

... which from my understanding can happen if the earlier call to `allocator->AllocateMemory` fails. In this case, `pAllocation` will be freed and explicitly set to `0` (for example [here](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/05973d8aeb1a4d12f59aadfb86d20decadba82d1/include/vk_mem_alloc.h#L13722)). Depending on the exact configuration and provided flags, the code may then attempt to dereference the null `pAllocation` in one of the three function calls below line 16163.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -16404,7 +16404,7 @@ VMA_CALL_PRE VkResult VMA_CALL_POST vmaCreateImage(
         pImageCreateInfo,
         allocator->GetAllocationCallbacks(),
         pImage);
-    if(res >= 0)
+    if(res == VK_SUCCESS)
     {
         VmaSuballocationType suballocType = pImageCreateInfo->tiling == VK_IMAGE_TILING_OPTIMAL ?
             VMA_SUBALLOCATION_TYPE_IMAGE_OPTIMAL :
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for this fix. I merged it.

However, it is not true that `res >= 0` will always evaluate to true. Please see the documentation of `VkResult` https://docs.vulkan.org/spec/latest/chapters/fundamentals.html#fundamentals-returncodes Error codes have negative values, `VK_SUCCESS` is zero, and positive values are reserved for other special-purpose but non-error results like `VK_INCOMPLETE`.

crud89:
Thanks! Yes, you're right - I guess I've somehow assumed they are unsigned. But I still think that discarding everything that's not `VK_SUCCESS` is the intended behavior and it should fix the clang-tidy complaints. 🙂

PR #460, opened 2025-01-25
Remove redundant if block

PawelGribov:
Leave only else part. If I understand this TLSF logic correctly, two adjacent free blocks are merged immediately when one of them is freed. So the previous physical block will never be free.


adam-sawicki-a:
Does this change fix any bug, or just removes the code that we suspect to never execute?

By the way, I fixed the triple "s" typo that you noticed.

adam-sawicki-a:
I am not sure either 🙂 and I don't have time right now to analyze it, as this is not a bug to be fixed, so I propose to leave it as-is.

PawelGribov:
ok, closed 🙂

PR #461, opened 2025-01-26
Add `vmaSetAllocationNameWithLength` for non-null-terminated strings and substrings

justinasvd:
This pull request introduces the new function `vmaSetAllocationNameWithLength`, which extends the functionality of `vmaSetAllocationName` by allowing allocation names to be set from strings that are not null-terminated. The function provides more flexibility when working with fixed-length strings or substrings.

In particular, this enables setting names from `std::string_view` or other non-null-terminated string sources. The introduced function could also be viewed as a marginally more efficient way to set allocation names when the length of a string is known beforehand:
```C++
// Example comparison:
vmaSetAllocationName(alloc, allocation, s.c_str());       // Null-terminated string
vmaSetAllocationNameWithLength(alloc, allocation, s.data(), s.length()); // Known length
// Here, `s` is an `std::string` object.
```

Backward Compatibility:
- No changes to the existing `vmaSetAllocationName` behavior.
- Existing code using `vmaSetAllocationName` remains fully functional.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -2070,6 +2070,21 @@ VMA_CALL_PRE void VMA_CALL_POST vmaSetAllocationName(
     VmaAllocation VMA_NOT_NULL allocation,
     const char* VMA_NULLABLE pName);
 
+/** \brief Sets pName in given allocation to new value.
+
+If `pName` is null, `len` must be zero. Otherwise, `pName` must point to a 
+valid string. The function makes a local copy of the first `len` characters from 
+the string pointed to by `pName` and sets it as the allocation's `pName`.
+The caller must ensure the string is valid for at least `len` bytes. String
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for this proposal. I am sorry but I don't want to include it into the library code.

In general, I agree with you that it is good to use explicit string length. I am personally an opponent of null-terminated strings. If only VMA had an interface in modern C++ (where we could use `std::string_view`) or any other programming language, strings would use an explicit length.

However, VMA has an interface in C similar to Vulkan. In Vulkan API, all the strings are null-terminated and don't use explicit length. Examples include `VkDebugMarkerMarkerInfoEXT::pMarkerName` and `VkApplicationInfo::pApplicationName`. Thus, in the name of avoiding feature creep, I don't want to extend VMA with this new function.

PR #463, opened 2025-02-28
Fix various memory safety issues

hrdisaac:
- Fixed implicit type promotion in VkDeviceSize calculation by explicitly casting SMALL_BUFFER_SIZE and shift operations to prevent sign extension.

- Ensured null termination in VmaStringBuilder::Add by resizing the buffer and appending '\0' after copying.

- Initialized class member m_hParentPool in the VmaDeviceMemoryBlock constructor to prevent potential undefined behavior.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -5651,8 +5651,9 @@ void VmaStringBuilder::Add(const char* pStr)
     if (strLen > 0)
     {
         const size_t oldCount = m_Data.size();
-        m_Data.resize(oldCount + strLen);
+        m_Data.resize(oldCount + strLen + 1);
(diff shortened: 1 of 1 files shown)


adam-sawicki-a:
Thank you for these suggestions. However, I don't think any of them are needed.

`VmaDeviceMemoryBlock::m_hParentPool` is unconditionally initialized at the beginning of the `Init` method. But OK, I added the initialization to the constructor as you suggested.

What makes you think that `VmaStringBuilder::m_Data` should be null-terminated? The data is only used as an input to the function `VmaCreateStringCopy`, which takes explicit length and adds null terminator to the created copy.

Did you meet any bugs related to the changes you proposed? Or were they suggested by some automatic or AI code analysis tools?

Issue #467, opened 2025-03-13
Use GitHub actions for continuous integration


IAmNotHanni:
Hello

This requires **Linux support** (https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/466)

Using [GitHub actions](https://github.com/features/actions) for Continuous Integration (CI) on GitHub has become more or less the industry standard. From my personal experience, there are many advantages of this approach. The CI is directly embedded into the GitHub repo, and GitHub is very generous with providing resource to run your CI for free.

My plan would be to add a simple CI workflow which compiles both debug and release for Windows and Linux. On Linux, we can use gcc and cmake, and for Windows, we can use MSVC compiler (and even clang for Windows). You can mark GitHub workflows as required, so pull requests are only merged if builds pass. This is very convenient.

In addition, I also would like to add a separate setup for static code analysis using clang-tidy. I would not recommend to mark this as required though, because it would certainly require a lot of cleanup in the existing code. Clang-tidy can be very detailed, and not ever warning makes sense for the project. Still, it is nice to have a look at the static code analysis from time to time, especially when new pull requests come in. Although it is a lot of work, I have managed to get to (almost) zero warnings in some of my projects, which resulted in very high code quality (although it was a lot of work and in some parts probably not worth it).

Also, I would recommend to store the results of the builds as artefacts in a nightly build. This allows you and others to download latest builds, and to run them directly.

While all of this sounds like a lot of code, maintaining a CI using GitHub actions has become very easy, and it's only a small number of GitHub workflow files to add. To see all of this in actions, check out our code here for example: 

https://github.com/inexorgame/vulkan-renderer/tree/main/.github/workflows
https://github.com/inexorgame/vulkan-renderer/actions/runs/13732338625/job/38411491197

Some more notes:

* I think it's important to check if both debug and release build work. I have experienced several code bases where release is totally broken because it is never really tested for.
* You could also migrate to https://github.com/google/googletest and integrate tests into the CI as well. I am personally not yet doing this either, because I also don't have a solution yet how to run code which requires a gpu in CI. There seem to be some commercial services for this, but I don't think GitHub offers this for free yet. It would be a great thing to suggest to Microsoft.
* Documentation could also be built in the CI. I can also recommend this a lot.

best regards
Johannes

adam-sawicki-a:
Hello

Thank you for this effort and I'm sorry for my delayed response.

[...]

PR #469, opened 2025-03-17
Add GitHub workflow for continuous integration for Windows

IAmNotHanni:
Related to https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/467

We can already use [GitHub actions](https://github.com/features/actions) for continuous integration (CI) for the Windows build.
This workflow builds the code using MSVC compiler and clang for Windows, each in both debug and release configuration.
The results are uploaded as artifacts in a nightly build.

Let me know if you request changes to the workflow.

best regards
Johannes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,117 @@
+name: Build Code
+
+on:
+  push:
+  pull_request:
+    types: [opened]
+
+env:
+  VMA_VULKAN_VERSION: "1.3.283.0"
+  VMA_VULKAN_SDK_PATH: "$GITHUB_WORKSPACE/../vulkan_sdk/"
+  VMA_VULKAN_SDK_CHECKSUM_LINUX: "8005e2cf3e89c80cbe1c0d0a259c88248de3257b4fc6fdefb47409edb3e43ecb"
+  VMA_VULKAN_SDK_CHECKSUM_WINDOWS: "811fcb9b43d09248520b2f38ae9a3763fc81df950fdab874f23bd762b07a9b12"
+
+jobs:
+  windows:
+    name: ${{ matrix.config.name }}
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {
+            name: "Windows MSVC (Debug)",
+            compiler: "msvc",
+            cc: "cl", cxx: "cl",
+            cmake_configure_options: '-G "Visual Studio 17 2022" -A x64',
+            build_type: "Debug",
+            cmake_build_options: "--config Debug",
+          }
+          - {
+            name: "Windows MSVC (Release)",
+            compiler: "msvc",
+            cc: "cl", cxx: "cl",
+            cmake_configure_options: '-G "Visual Studio 17 2022" -A x64',
+            build_type: "Release",
+            cmake_build_options: "--config Release",
+          }
+          - {
+            name: "Windows Clang (Debug)",
+            compiler: "clang",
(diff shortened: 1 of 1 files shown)


IAmNotHanni:
To test the workflow, I think you must first agree to run it.

.github/workflows/build.yml
```diff
@@ -0,0 +1,117 @@
...
+
+on:
+  push:
+  pull_request:
+    types: [opened]
+
+env:
+  VMA_VULKAN_VERSION: "1.3.283.0"
+  VMA_VULKAN_SDK_PATH: "$GITHUB_WORKSPACE/../vulkan_sdk/"
+  VMA_VULKAN_SDK_CHECKSUM_LINUX: "8005e2cf3e89c80cbe1c0d0a259c88248de3257b4fc6fdefb47409edb3e43ecb"
```
adam-sawicki-a, line 11: These checksums don't seem to be used. Can we get rid of them? Even if you could use them, I would rather avoid having to update such checksums every time we update Vulkan SDK.

adam-sawicki-a (changes requested):
Just one comment, otherwise looks good to me.

adam-sawicki-a:
I noticed that a downloadable artifact is a ZIP archive with another ZIP archive inside. Can we avoid this double-packing?

adam-sawicki-a:
I can see you are downloading LLVM from repository https://github.com/zufuliu/llvm-utils. It doesn't look like an official LLVM account. Is there no official distribution of those binaries?

IAmNotHanni:
> I noticed that a downloadable artifact is a ZIP archive with another ZIP archive inside. Can we avoid this double-packing?
> 
> I am not sure if we need any artifacts to download later. This is a library distributed in the source form. All we need from the automated builds is the log with errors and warnings.

Oh, true. I will remove this and see if it still compiles (if clang is installed by default in visual studio in the workflow).
If not, I will just remove clang for Windows for now. It's not that important I think, and we still have clang for Linux.

adam-sawicki-a:
It seems that Clang has failed.

> C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(459,5): error MSB8020: The build tools for LLVM_v143 (Platform Toolset = 'LLVM_v143') cannot be found. To build using the LLVM_v143 build tools, please install LLVM_v143 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [D:\a\VulkanMemoryAllocator\VulkanMemoryAllocator\build\CMakeFiles\3.31.6\VCTargetsPath.vcxproj]

IAmNotHanni:
> It seems that Clang has failed.
> 
> > C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(459,5): error MSB8020: The build tools for LLVM_v143 (Platform Toolset = 'LLVM_v143') cannot be found. To build using the LLVM_v143 build tools, please install LLVM_v143 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [D:\a\VulkanMemoryAllocator\VulkanMemoryAllocator\build\CMakeFiles\3.31.6\VCTargetsPath.vcxproj]

https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/473 :D

PR #473, opened 2025-03-28
Remove unrequired Windows clang build in GitHub

IAmNotHanni:
One more thing about https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/469:
I removed windows clang build in CI because it's not working for now.
Maybe I can find a way to use clang for Windows using some offical repositories.


adam-sawicki-a:
OK, thanks. If we just test Clang on Linux, this is fine. I guess errors and warnings from this compiler should be similar on any platform.

Issue #433, opened 2024-07-10
Write a new test which covers 'advanced data uploading' from docs


IAmNotHanni:
In my opinion, it would be better to have a test which covers the topic of advanced data uploading in detail. The current sample code from the docs still has the following issues:

* It's commented code, so it's not really run (compiler doesn't even check if syntax is valid for example)
* The example from the docs only covers the use case of a uniform buffer

I think programmers could learn a lot a test that covers most common cases (vertex, index, and uniform buffer maybe?), especially when it comes to correct barrier placement. The benefit of such a test would be that we can run it with synchronization validation layers to ensure the barriers are correct. This would give new programmers a good, safe code to use as reference.

I propose to add the following test in `Tests.cpp`:

```cpp
static void TestAdvancedDataUploading() {
    wprintf(L"Testing advanced data uploading\n");

    auto create_buffer = [](VkDeviceSize bufferSize, VkBufferUsageFlags bufferUsage, VmaAllocationCreateFlags allocationFlags,
        VkBuffer& buffer /*out*/, VmaAllocation& alloc /*out*/, VmaAllocationInfo& allocInfo /*out*/, void* myData,
        std::size_t myDataSize) {

        TEST(myData != nullptr);
        TEST(myDataSize != 0);

        VkBufferCreateInfo bufCreateInfo = { VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO };
        bufCreateInfo.size = bufferSize;
        bufCreateInfo.usage = bufferUsage;

        VmaAllocationCreateInfo allocCreateInfo = {};
        allocCreateInfo.usage = VMA_MEMORY_USAGE_AUTO;
        allocCreateInfo.flags = allocationFlags;

        VkResult result = vmaCreateBuffer(g_hAllocator, &bufCreateInfo, &allocCreateInfo, &buffer, &alloc, &allocInfo);
        TEST(result == VK_SUCCESS);

        VkMemoryPropertyFlags memPropFlags;
        vmaGetAllocationMemoryProperties(g_hAllocator, alloc, &memPropFlags);

        BeginSingleTimeCommands();

        if (memPropFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT) {
            // Allocation ended up in a mappable memory and is already mapped - write to it directly.
            memcpy(allocInfo.pMappedData, myData, myDataSize);
[...]

PR #474, opened 2025-03-29
Add new tests for advanced data uploading

IAmNotHanni:
Closes https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/433

What do you think? The new tests don't show any validation layer warnigns or errors on my system.
I still need to find a way to link the code in the doxygen documentation, but that's for another pr.

I still have some **questions**:

1) In `TestDataUploadingWithStagingBuffer`, I explicitely check if the memory did not end up in mappable memory with `TEST(!(memPropFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT));`. Does this make sense like this? Because if this is not true, it would be a bug in vma, right? (The same argument is for `TEST(memPropFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT);` in `TestDataUploadingWithMappedMemory`).

2) Besides `vmaFlushAllocation`, should we maybe also mention in a comment that the user needs to call `vmaInvalidateAllocation` if reading from gpu -> cpu is the desired use case?

3) In theory, we could abstract `TestAdvancedDataUploading` into a function (as I did with the [lambda in the ticket here](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/433)) so we have less duplicate code. The functions `TestDataUploadingWithMappedMemory` and `TestDataUploadingWithStagingBuffer` would then call that function with the flags required. While this leads to less code, I am not sure if it would be easier for beginners to understand.

4) Also, I think it would be worth mentioning in the docs that both `vmaFlushAllocation` and `vmaInvalidateAllocation` check internally if the memory is `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT`, meaning we don't need to check for manually this when calling these functions

5) Do we need some additional tests if the data uploading was successful?

Johannes

### DIFF
--- a/src/Tests.cpp
+++ b/src/Tests.cpp
@@ -6161,7 +6161,7 @@ static void TestDataUploadingWithStagingBuffer()
     const VkDeviceSize bufferSize = 65536;
     std::vector<std::uint8_t> bufferData(bufferSize);
     for (auto& bufferByte : bufferData) {
-        bufferByte = static_cast<std::uint8_t>(rand() % 256);
+        bufferByte = static_cast<std::uint8_t>(rand());
     }
 
     VkBufferCreateInfo uniformBufferCI = { VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO };
@@ -6170,19 +6170,13 @@ static void TestDataUploadingWithStagingBuffer()
 
     VmaAllocationCreateInfo uniformBufferAllocCI = {};
     uniformBufferAllocCI.usage = VMA_MEMORY_USAGE_AUTO;
-    uniformBufferAllocCI.flags = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
 
     VkBuffer uniformBuffer = VK_NULL_HANDLE;
-    VmaAllocation uniformBufferAlloc = {};
+    VmaAllocation uniformBufferAlloc = VK_NULL_HANDLE;
     VmaAllocationInfo uniformBufferAllocInfo = {};
 
     VkResult result = vmaCreateBuffer(g_hAllocator, &uniformBufferCI, &uniformBufferAllocCI, &uniformBuffer, &uniformBufferAlloc, &uniformBufferAllocInfo);
     TEST(result == VK_SUCCESS);
-
-    // We need to check if the uniform buffer really ended NOT up in mappable memory.
-    VkMemoryPropertyFlags memPropFlags;
-    vmaGetAllocationMemoryProperties(g_hAllocator, uniformBufferAlloc, &memPropFlags);
-    TEST(!(memPropFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT));
 
     VkBufferCreateInfo stagingBufferCI = { VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO };
     stagingBufferCI.size = bufferSize;
@@ -6200,10 +6194,7 @@ static void TestDataUploadingWithStagingBuffer()
     TEST(result == VK_SUCCESS);
 
     TEST(stagingBufferAllocInfo.pMappedData != nullptr);
-    std::memcpy(stagingBufferAllocInfo.pMappedData, bufferData.data(), bufferData.size());
-
-    result = vmaFlushAllocation(g_hAllocator, uniformBufferAlloc, 0, VK_WHOLE_SIZE);
-    TEST(result == VK_SUCCESS);
+    vmaCopyMemoryToAllocation(g_hAllocator, bufferData.data(), stagingBufferAlloc, 0, bufferData.size());
 
     BeginSingleTimeCommands();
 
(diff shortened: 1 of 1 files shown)


src/Tests.cpp
```diff
@@ -6153,6 +6153,264 @@ static void TestMemoryUsage()
...
 
+static void TestDataUploadingWithStagingBuffer()
+{
+    wprintf(L"Testing data uploading with staging buffer...\n");
+
+    // Generate some random data to fill the uniform buffer with.
+    const VkDeviceSize bufferSize = 65536;
+    std::vector<std::uint8_t> bufferData(bufferSize);
+    for (auto& bufferByte : bufferData) {
+        bufferByte = static_cast<std::uint8_t>(rand() % 256);
```
adam-sawicki-a, line 6164: When you cast to `uint8_t`, the value gets modulo 256, so explicit modulo is redundant.

src/Tests.cpp
```diff
@@ -6153,6 +6153,264 @@ static void TestMemoryUsage()
...
+    }
+
+    VkBufferCreateInfo uniformBufferCI = { VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO };
+    uniformBufferCI.size = bufferSize;
+    uniformBufferCI.usage = VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT; // Change this if you want to create another type of buffer.
+
+    VmaAllocationCreateInfo uniformBufferAllocCI = {};
+    uniformBufferAllocCI.usage = VMA_MEMORY_USAGE_AUTO;
+    uniformBufferAllocCI.flags = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
```
adam-sawicki-a, line 6173: 🐞 `VmaAllocationCreateInfo::flags` is not intended to use `VK_MEMORY_PROPERTY_` flags there!

This is just not needed. There is no reason to force the allocation to be in `DEVICE_LOCAL` memory.

src/Tests.cpp
```diff
@@ -6153,6 +6153,264 @@ static void TestMemoryUsage()
...
+
+    VkBuffer uniformBuffer = VK_NULL_HANDLE;
+    VmaAllocation uniformBufferAlloc = {};
```
adam-sawicki-a, line 6176: Just as I already mentioned in #433:

>You initialize one with `VK_NULL_HANDLE` but another one with default initialization. I recommend to be consistent and to use `VK_NULL_HANDLE` everywhere, like it is done in other places of the code.

src/Tests.cpp
```diff
@@ -6153,6 +6153,264 @@ static void TestMemoryUsage()
...
+    VmaAllocationInfo uniformBufferAllocInfo = {};
+
+    VkResult result = vmaCreateBuffer(g_hAllocator, &uniformBufferCI, &uniformBufferAllocCI, &uniformBuffer, &uniformBufferAlloc, &uniformBufferAllocInfo);
+    TEST(result == VK_SUCCESS);
+
+    // We need to check if the uniform buffer really ended NOT up in mappable memory.
```
adam-sawicki-a, line 6182: Why do you think this check is needed? I think you shouldn't check that. What if we have a platform with unified memory (or just ReBAR enabled) where every memory type is `HOST_VISIBLE`? It is perfectly valid, while it would fail your check.

src/Tests.cpp
```diff
@@ -6153,6 +6153,264 @@ static void TestMemoryUsage()
...
+        bufferByte = static_cast<std::uint8_t>(rand() % 256);
+    }
+
+    VkBufferCreateInfo uniformBufferCI = { VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO };
+    uniformBufferCI.size = bufferSize;
+    uniformBufferCI.usage = VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT; // Change this if you want to create another type of buffer.
+
+    VmaAllocationCreateInfo uniformBufferAllocCI = {};
+    uniformBufferAllocCI.usage = VMA_MEMORY_USAGE_AUTO;
+    uniformBufferAllocCI.flags = VMA_ALLOCATION_CREATE_HOST_ACCESS_SEQUENTIAL_WRITE_BIT | VMA_ALLOCATION_CREATE_HOST_ACCESS_ALLOW_TRANSFER_INSTEAD_BIT | VMA_ALLOCATION_CREATE_MAPPED_BIT;
```
adam-sawicki-a, line 6316: This line is so long I advise breaking it into at least 2.

adam-sawicki-a (changes requested):
Thank you for adding these tests. I'm happy to merge when the issues I posted as comments are fixed.

adam-sawicki-a:

PR #478, opened 2025-04-03
Enable all gcc and clang warnings for testing the CI

IAmNotHanni:
* Some of these warnings are probably way too overpowered, but we can remove them as necessary.
* None of these warnings should cause the build to fail (specifying warnings as errors `-Werror` would do this)


adam-sawicki-a:
Build failed after merging this change. Possibly Cmake syntax is broken.

PR #482, opened 2025-04-09
Add workflow for static code analysis using clang-tidy

IAmNotHanni:
* Use clang-tidy to perform static code analysis 
* Ignore warnings `-modernize-use-trailing-return-type`, `cppcoreguidelines-macro-usage`, `modernize-use-auto`

clang-tidy is very powerful, but it needs proper configuration because you sometimes disagree with some things it comes up with. Let me know if you have further warnings to add to the ignore list. Still, I find it interesting what it already can find.

### DIFF
--- a/.github/workflows/static_code_analysis.yaml
+++ b/.github/workflows/static_code_analysis.yaml
@@ -1,0 +1,45 @@
+name: Static code analysis (clang-tidy) 
+
+on:
+  push:
+  pull_request:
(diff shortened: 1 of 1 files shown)


IAmNotHanni:
Seems to work: https://github.com/IAmNotHanni/VulkanMemoryAllocator/actions/runs/14368258430/job/40286113078

Issue #484, opened 2025-04-10
Add some warnings from clang-tidy to ignore list


IAmNotHanni:
Hello

I think some warnings from clang-tidy can be added to the ignore list, since in my opinion they are either not that helpful or out of the scope of the project.
If you have any to add, let me know.

best regards
Johannes

IAmNotHanni:
Here is a list of warnigns I personally would ignore:

* `altera-unroll-loops` ("kernel performance could be improved by unrolling this loop" I don't think this is necessary)
* `misc-definitions-in-headers` (I think definitions in header file are fine for this project)
* `cppcoreguidelines-pro-bounds-pointer-arithmetic` (I think this project just relies on pointer arithmetics)
* `readability-function-cognitive-complexity` (Some functions are just long)
...

IAmNotHanni:
@adam-sawicki-a I just saw your comment on https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/482
I enabled all warnings by default, and now you can pick which ones you want to ignore. (I don't know all warnings myself)

adam-sawicki-a:
OK, your list looks good to me. If you add those I mentioned in that comment https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/482#issuecomment-2792412996, that will be great.

IAmNotHanni:
About the analysis of `CMakeCXXCompilerId.cpp`: I need to check how to configure `clang-tidy` more correctly so it only analyzes the files of VMA. I'm looking into this.

adam-sawicki-a:
Maybe there is a way to tell clang-tidy the version of C++ we are targeting (C++14)?

IAmNotHanni:
> Maybe there is a way to tell clang-tidy the version of C++ we are targeting (C++14)?

Just for my understanding: Does VMA really not use any C++17 feature?
Because if it's really C++14 only, there is this:

https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/fc2cea529d799e15815a54882a4a2fa646c014d2/src/CMakeLists.txt#L37

IAmNotHanni:
I also think `llvmlibc-implementation-in-namespace` makes not much sense.

IAmNotHanni:
[...]

PR #489, opened 2025-04-11
Specify C++14 as standard in CMake, not C++17 ?

IAmNotHanni:
I noticed this in this discussion: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/484#issuecomment-2794352480

Does this make sense? This is a C++14 project, right?
If we're not using C++17, we can specify C++14 as intended.
Maybe this was overlooked when CMake was introduced to VMA.


adam-sawicki-a:

PR #491, opened 2025-05-09
Use a cache instead of downloading Vulkan SDK with every CI run

### LINKED ISSUE #490: Cache Vulkan SDK in GitHub workflow
* We don't need to download the Vulkan SDK every time. If we cache it, this step can be skipped in build and clang-tidy GitHub workflows.
* Sadly, there seems to be no such easy fix for the apt packages (other than using a custom Docker image or custom runner).

IAmNotHanni:
Closes #490 

On the second run, the step of downloading Vulkan SDK should be skipped in build (windows and linux) as well as in clang-tidy.
Here's an example of Windows MSVC debug build:

grafik

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -1,4 +1,4 @@
-name: Build code on Linux
+name: Build code
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


adam-sawicki-a:

PR #494, opened 2025-05-18
Small improvement of 'advanced data uploading' exampe in docs

IAmNotHanni:
`vmaCopyMemoryToAllocation` is used in the tests, so it should be used in the docs as well.


IAmNotHanni:
Should we add a note that the user should not use `vmaCopyMemoryToAllocation` if it's the intention to keep the memory persistently mapped? I was a little surprised that `vmaCopyMemoryToAllocation` actually does the unmap, but I think it's safer to do it this way instead of assuming the programmer wants it to stay mapped?

adam-sawicki-a:

PR #496, opened 2025-05-21
Small improvements of tests

IAmNotHanni:
`vmaFlushAllocation` is unnecessary here because [vmaCopyMemoryToAllocation](https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/group__group__alloc.html#ga11731ec58a3a43a22bb925e0780ef405) does that internally if it's required.
This small mistake made it somehow through when discussing pull request https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/474.
I only found this in this test, the others seem to be correct.


IAmNotHanni:
Oh I just also realized the tests I wrote are not checking the return value of `vmaCopyMemoryToAllocation`, but they really should, I will add this in another pr.

Issue #498, opened 2025-06-13
Document error codes


Majora320:
It would be useful to have documentation about how different functions can fail, similar to how the Vulkan spec lists possible error codes for each operation. This should be possible by unioning the possible error codes from called Vulkan functions for each VMA function, and combining them with whatever Vulkan error codes VMA uses itself, maybe with a little bit of documentation for the latter case. Not a high priority issue, but it would be nice to have.

sawickiap:
Thank you for the suggestion. Can you please explain how such list of error codes included in the documentation of each function could be used in practice?

I don't think is is very useful, while creating and maintaining it would require a lot of work.

Whenever a specific returned value has a specific meaning (like in `vmaFindMemoryTypeIndex`), it is documented. Otherwise, it should be assumed that every code other than `VK_SUCCESS` indicates "an error" (as in - the operation failed).

Vulkan and DX12 functions are known to return error codes beyond the official list anyway, so even graphics drivers don't always obey this specification.

PR #510, opened 2025-10-15
Add VMA_RESULT for possible success codes

YaaZ:
A new macro is used to document possible success codes for functions.
This may be useful just as a form of documentation, but even more useful for projects like [VulkanMemoryAllocator-Hpp](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp), which uses the VMA header to generate bindings, taking into account success codes when throwing exceptions.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -273,6 +273,13 @@ extern "C" {
     #define VMA_EXTENDS_VK_STRUCT(vkStruct)
 #endif
 
+// Define this macro to decorate the return type with possible success return codes.
+// Undecorated functions are assumed to return VK_SUCCESS on success.
+// Error codes are not included there.
+#ifndef VMA_RESULT()
+    #define VMA_RESULT(...)
+#endif
+
 // Define this macro to decorate pointers with an attribute specifying the
 // length of the array they point to if they are not null.
 //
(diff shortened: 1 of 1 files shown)


sawickiap:
I am sorry, but this proposal has been discussed and rejected already as #498.

YaaZ:
Actually, I agree with your comment in #498, but my proposal is different in that it only includes success codes, not every code. With the same concerns:

> creating and maintaining it would require a lot of work.

Not with the success codes.

> Vulkan and DX12 functions are known to return error codes beyond the official list anyway

Should not be relevant if we limit the scope to success codes, right?

> Whenever a specific returned value has a specific meaning (like in vmaFindMemoryTypeIndex), it is documented.

It's documented in a free form, what I propose is a strict format.

Interesting thing by the way: Vulkan-Hpp throws exceptions for error codes and VMA-Hpp currently does the same, but for `vmaFindMemoryTypeIndex` and similar functions, `VK_ERROR_FEATURE_NOT_PRESENT` is one of the normal results - it doesn't look to me like we should throw exception there. Although the result code indicates an error, semantically it's a status code.
So you've chosen the most suitable code you could find among existing codes, but did you originally consider this an error, or status code?

It seems, such confusion is a sign, that strict documentation of return codes could indeed be helpful, but if you reject my proposal, the discussion is still valuable in itself.

sawickiap:
I am sorry but VMA is a C++ library with interface in C, while any binding to other programming languages are out of scope of this repository. To avoid feature creep, I don't want to make the code longer or more convoluted to support such bindings.

YaaZ:
I get it, thank you. Could you answer my question about `VK_ERROR_FEATURE_NOT_PRESENT`, please?
Do you consider this a status or error code in the context of `vmaFindMemoryTypeIndex`?
I am confused because semantically it looks like a status code - a normal result for a query function to find nothing.
On the other hand, the constant itself belongs to the error group (VK_ERROR prefix & negative value).
Also, following the Vulkan spec, returning error code leaves output parameters undefined, which is indeed the case with `vmaFindMemoryTypeIndex` (it may either leave it unchanged, or set to `UINT32_MAX`), so behavior-wise it seems to lean more to the "error" group.

sawickiap:
Yes, as you said, I also think that semantically returning `VK_ERROR_FEATURE_NOT_PRESENT` makes sense because we can say the function formally failed to find a memory type and the output parameter `*pMemoryTypeIndex` is then undefined.

Whether this situation deserves to be a C++ exception is a problem of C++ and, as such, it is out of scope of the VMA repository, which defines only interface in C and doesn't use exceptions, even internally.

In other programming languages, which can express the intent "returned uint memory type index or error code" more directly, like Rust, it could be something like:

```rust
fn FindMemoryTypeIndex(...) -> Result<u32, VkError> { ... }
```

YaaZ:
That makes sense, thank you for your answers.

PR #512, opened 2025-10-16
Adjusted doc wording for pTypeExternalMemoryHandleTypes

### LINKED ISSUE #508: Cannot use memory auto-export + import at the same time
VMA version: 3442720aca62575eabbcd9d1b5926113360717ee

What i'm trying to do is to have all VMA-allocated GPU images/buffers be exportable by default while also supporting memory import (auto-export and import are mutually exclusive here).

For the auto-export part, I set the supported handle types in `VmaAllocatorCreateInfo::pTypeExternalMemoryHandleTypes` for all `DEVICE_LOCAL` memory types.

For the import part, I use `vmaCreateDedicatedImage`/`vmaCreateDedicatedBuffer` + `VkImportMemoryWin32HandleInfoKHR` + custom pool, since the doc says that `VmaAllocatorCreateInfo::pTypeExternalMemoryHandleTypes` only affects default pools.

The problem I'm facing is that when I import memory, `VkExportMemoryAllocateInfoKHR` from the auto-export part still gets pushed to the `VkMemoryAllocateInfo` pNext chain, which means I end up with both `VkExportMemoryAllocateInfoKHR` with export handle type + `VkImportMemoryWin32HandleInfoKHR` with import handle type.

As the handle types I use for auto-export and import are not compatible with each other (`D3D11_TEXTURE` for import and `OPAQUE_WIN32` for export), this doesn't work.

maxime-modulopi:
Closes #508


sawickiap:
Thanks for this suggestion. HTML documentation is generated by Doxygen based on special comments in "vk_mem_alloc.h" and shouldn't be modified directly. But it doesn't matter. I applied your suggested change in the right place (for now on my fork of the repository - https://github.com/sawickiap/VulkanMemoryAllocator - it will be merged to the upstream later).

PR #514, opened 2025-10-24
Fixing symbol exports for C++20 module

### LINKED ISSUE #513: Fails to build with clang 21 when used via C++ module
All calls to `VmaFree` (at least) in the `vk_mem_alloc.h` header fails with:
```
error: no matching function call to 'VmaFree'
```

There is an issue open in the LLVM project, not sure who is to blame or if there's some workaround, or if it will just be broken until new clang version has been released, https://github.com/llvm/llvm-project/issues/157327

This surfaces when used as a C++ module, for example via https://github.com/YaaZ/VulkanMemoryAllocator-Hpp.

M2-TE:
This resolves #513.

Some background first:
When using C++20 modules, exporting VMA leads to issues with the newly released Clang 21. It e.g. complains about `VmaFree` being an unkown symbol, as it is a TU-local entity with internal linkage (`static`), which cannot be exported. The solution here is to also mark the function as `inline`, which allows the compiler to export this function as part of another function that _can_ be exported.

Looking through the rest of the code, there are a few redundant or contradictory uses of `inline` and `static`, which I wanted to hear some thoughts about:

- Templated functions: These will always be inlined, so adding `inline` is redundant, but might be nice as an indication of intent. Some of these are also marked as `static` however, which would only really make sense for template specializations (of which there aren't any, as far as I could tell).
- `static` in general: This library is intended to be used in both C and C++ environments, correct? I only have experience with the latter, so correct me if it is wrong for either, but do static functions even make sense here (inside the implementation portion)? The header alongside `VMA_IMPLEMENTATION` should only be used in a single location in the project, so there won't be any ODR violations. The `static` keyword for this purpose was supposed to be deprecated in C++11 in favor of using anonymous namespaces instead. Since the implementation side is C++, that might be a better solution, if this functionality is even desired.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -3552,6 +3552,7 @@ constexpr uint32_t VMA_VENDOR_ID_AMD = 4098;
 // Vulkan 1.0, but doesn't actually define it in Vulkan SDK earlier than 1.2.131.
 // See pull request #207.
 #define VK_ERROR_UNKNOWN_COPY ((VkResult)-13)
+} // namespace
 
 
 #if VMA_STATS_STRING_ENABLED
@@ -3602,11 +3603,11 @@ enum class VmaAllocationRequestType
 
 #endif // _VMA_ENUM_DECLARATIONS
 
-} // namespace
-
 #ifndef _VMA_FORWARD_DECLARATIONS
 // Opaque handle used by allocation algorithms to identify single allocation in any conforming way.
 VK_DEFINE_NON_DISPATCHABLE_HANDLE(VmaAllocHandle);
+
+struct VmaBufferImageUsage;
 
 struct VmaMutexLock;
 struct VmaMutexLockRead;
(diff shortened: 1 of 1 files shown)


sawickiap:
I've pushed the fix as you suggested, on my fork for now:

Please let me know if it works for you.

sawickiap:
Thank you for these suggestions. I think you are right - `inline` and `static` may be used inconsistently throughout the library source code. My intent was that:

- `inline` is used whenever a function is simple, to hint the compiler to prefer inlining it.
- `static` is used for global functions, constants, variables to make them local to the .cpp file so they don't participate in the linking process, which might have some performance cost and could generate errors if clashing with a function with the same name in another .cpp file.

In the latter case, I agree that using an anonymous namespace is a better solution, considering the implementation of VMA is in C++14. I've changed in. Please let me know if it works for you.

M2-TE:
Your fork works for me; tested with Clang 21 and compiled as part of a C++20 module.

Anonymous namespace implementation LGTM! Marking some `static const` as `constexpr` inside it was also a good idea.
You could likely do the same to some of the `static const` members in your structs/classes, e.g. `VmaMappingHysteresis`:

```cpp
class VmaMappingHysteresis
{
    // static const int32_t COUNTER_MIN_EXTRA_MAPPING = 7; // C-style
    static constexpr int32_t COUNTER_MIN_EXTRA_MAPPING = 7; // C++-style
};
```

Marking them as `static constexpr` avoids any storage for these inside the structs, just like the current `static const`. It is a stronger hint to the compiler, that these can (and should) be evaluated at compile time for some extra potential optimizations.

sawickiap:
Thanks for this suggestion. I changed some member variables from `static const` to `static constexpr`.

M2-TE:
Looks good!

M2-TE:
I've fixed the issue about export of TU-local entities with `GCC15` that I talked about in https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/513#issuecomment-3562367932 so please give it a quick read @sawickiap. Short summary of changes:
- `VmaBufferImageUsage` was moved outside the anonymous namespace as it is used in parameters. It was also moved before `_VMA_FUNCTIONS`, since `FindMemoryPreferences` makes use of it.
- Split declaration and definition for `VmaVector::~VmaVector()`
- Split declaration and definition for `VmaStlAllocator::allocate` and `VmaStlAllocator::deallocate`

sawickiap:
Thank you for these fixes. I applied them on my fork.

PR #522, opened 2026-01-19
CMake: Additional target ALIAS in the VulkanMemoryAllocator namespace

M2-TE:
As per title, this simply adds the new alias `VulkanMemoryAllocator::Headers`.

The goal is to have a sort of parallel to the `Vulkan::Headers`, `Vulkan::Hpp` and `Vulkan::HppModule` targets.
[YaaZ/VulkanMemoryAllocator-Hpp](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp) would then introduce `VulkanMemoryAllocator::Hpp` and `VulkanMemoryAllocator::HppModule` targets to be consistent with the Vulkan targets.

It does not break anything and would only really be a cosmetic change, as another ALIAS target already exists (`GPUOpen::VulkanMemoryAllocator`). Having the same consistent target relationship between VMA and its Hpp bindings, similar to Vulkan and Vulkan-Hpp bindings, would be great!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ project(VMA VERSION 3.3.0 LANGUAGES CXX)
 
 add_library(VulkanMemoryAllocator INTERFACE)
 add_library(GPUOpen::VulkanMemoryAllocator ALIAS VulkanMemoryAllocator)
+add_library(VulkanMemoryAllocator::Headers ALIAS VulkanMemoryAllocator)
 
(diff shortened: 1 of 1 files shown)


sawickiap:
Thank you for this improvement. I merged it into my fork for now:
It will be merged to the main repo in the future.

M2-TE:
Since this was closed, was the PR rejected or will it simply be brought in as part of the fork from @sawickiap as planned?

sawickiap:
It will be merged to the upstream repo soon.

PR #527, opened 2026-03-03
Added `VMA_MAYBE_UNUSED` macro for `[[maybe_unused]]` in C++17

M2-TE:
There are quite a few unused function parameters and member variables, which spit out numerous warnings during compilation. As an example with Clang 21:
```shell
vma/include/vk_mem_alloc.h:16694:48: warning: unused parameter 'allocator' [-Wunused-parameter]
16694 |     VmaAllocator VMA_NOT_NULL allocator,
                                      ^
```

Parameters are either unused due to only appearing in asserts, or because functions do not use those parameters at all. I reckon they are still there simply for the sake of ABI compatibility? Overloads probably aren't an option, since that only works for C++.

Either way, C++17 introduced the [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute, which silences warnings when variables end up unused. For compatibility, it is hidden behind the `VMA_MAYBE_UNUSED` macro, only evaluating to this attribute when compiling with C++17.

### DIFF
--- a/include/vk_mem_alloc.h
+++ b/include/vk_mem_alloc.h
@@ -328,6 +328,24 @@ extern "C" {
     #endif
 #endif
 
+#ifndef VMA_MAYBE_UNUSED
+    #if defined(_MSVC_LANG)
+        #if 201402L < _MSVC_LANG
+            #define VMA_MAYBE_UNUSED [[maybe_unused]]
+        #else
+            #define VMA_MAYBE_UNUSED
+        #endif
+    #elif defined(__cplusplus)
+        #if 201402L < __cplusplus
+            #define VMA_MAYBE_UNUSED [[maybe_unused]]
+        #else
+            #define VMA_MAYBE_UNUSED
+        #endif
+    #else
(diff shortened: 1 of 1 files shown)


M2-TE:
Just saw that there is `VMA_CPP20`. The whole language macros could be merged into a single `VMA_CPP`, containing the C++ version.

Also, I need to `#undef` the `VMA_MAYBE_UNUSED`, where would be the best spot for that?

sawickiap:
I am sorry, but the documentation clearly states in the "Features not supported" chapter:

>Code free of any compiler warnings. Maintaining the library to compile and work correctly on so many different platforms is hard enough. Being free of any warnings, on any version of any compiler, is simply not feasible. There are many preprocessor macros that make some variables unused, function parameters unreferenced, or conditional expressions constant in some configurations. The code of this library should not be bigger or more complicated just to silence these warnings. It is recommended to disable such warnings instead.

M2-TE:
Oh I see, missed that part.
Thanks for letting me know!

Issue #313, opened 2023-01-18
Possible race condition on VmaAllocator between defragmentation and regular allocations


sbourasse:
We've been using VMA extensively for our memory management on projects with heavy memory load. However, upon integrating the defragmentation utility, we started getting frequent memory corruption on internal VMA data. We narrowed down the source of this corruption to concurrent accesses on our `VmaAllocator` and (apparently) succeeded in preventing it by blocking allocations for the duration of the defragmentation. More context follows, with a couple questions at the end.

We use a single `VmaAllocator` from two separate threads. Most of our allocations are done from the rendering thread, which is also the one that runs the defragmentation. We also allocate from the main thread, although less frequently.

Our defragmentation integration looks something like this (it has the same structure as what can be found in the online documentation) :
```
vmaBeginDefragmentation(mainAllocator, ...);

for (;;)
{
	if (vmaBeginDefragmentationPass(mainAllocator, ...) == VK_SUCCESS)
	   break;

	CreateResourcesAndCopy();
	WaitForCompletion();

	if (vmaEndDefragmentationPass(mainAllocator, ...) == VK_SUCCESS)
		break;
}

vmaEndDefragmentation(mainAllocator, ...);
```

Running this bit of code on heavy workloads from our rendering thread while our main thread is free to use `mainAllocator` for other allocations, we reliably trigger memory corruptions, specifically located on the first 4 bytes of random `VmaAllocation` from our working `VmaDefragmentationPassMoveInfo::pMoves` array.

Regardless of this particular symptom, we managed to prevent such corruption by implementing a mutex targeting the region between `vmaBeginDefragmentation()` and `vmaEndDefragmentation()` as well as other VMA calls on `mainAllocator`, specifically those that could be made concurrently to defragmentation.

As a side note, we first prevented memory corruption by enabling `VMA_DEBUG_GLOBAL_MUTEX`, changing `VmaMutex::m_Mutex` type to `std::recursive_mutex`, and locking it from our side right after `vmaBeginDefragmentation()` and releasing it right before `vmaEndDefragmentation()`.

Does it make sense to treat the region from `vmaBeginDefragmentation()` to `vmaEndDefragmentation()` as a critical section with regards to the used `VmaAllocator` ? If so, is it something that could be enforced from within the library ?

adam-sawicki-a:
[...]

PR #529, opened 2026-03-12
Added missing pool locks in the VmaDefragmenterContext

JustSid:
We ran into the same issue that's also detailed in #313 and turol was nice enough to run this with TSAN so I figured I would be nice enough and add the missing locks.

There is a way to trigger this without running TSAN by using a debug sort function like MSVC provides:

```cpp
template<typename it, typename comp>
void debug_check_sort_fwd(it first, it last, comp c)
{
	assert(std::is_sorted(first, last, c));
}

template<typename RandomAccessIterator, typename Comp>
void debug_check_sort(RandomAccessIterator first_it, RandomAccessIterator last_it, Comp comp)
{
	using diff_t = typename iterator_traits<RandomAccessIterator>::difference_type;

	dev_assert(std::is_sorted(first_it, last_it, comp));
	// Limit the number of elements we need to check.
	diff_t size = last_it - first_it > diff_t(100) ? diff_t(100) : last_it - first_it;
	diff_t p    = 0;
	while (p < size) {
		diff_t q = p + diff_t(1);
		// Find first element that is greater than *(first_it+p).
		while (q < size && !comp(*(first_it + p), *(first_it + q))) {
			++q;
		}
		// Check that the elements from p to q are equal between each other.
		for (diff_t b = p; b < q; ++b) {
			for (diff_t a = p; a <= b; ++a) {
				assert(!comp(*(first_it + a), *(first_it + b)));
				assert(!comp(*(first_it + b), *(first_it + a)));
			}
		}
		// Check that elements between p and q are less than between q and size.
		for (diff_t a = p; a < q; ++a) {
			for (diff_t b = q; b < size; ++b) {
				assert(	comp(*(first_it + a), *(first_it + b)));
				assert(	!comp(*(first_it + b), *(first_it + a)));
			}
		}
		// Skip these equal elements.
		p = q;
	}
}

template<typename I>
void debug_sort(I first, I last)
{
	std::sort(first, last);
	debug_check_sort(first, last, std::less<decltype(*first)>());
}

template<typename I, typename C>
void debug_sort(I first, I last, C comp)
{
	std::sort(first, last, comp);
	debug_check_sort(first, last, comp);
}
```

If you slot this into `VMA_SORT()` you will occasionally hit the sort assertions depending on how the race goes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
   - Added function `vmaGetMemoryWin32Handle2` offering extra parameter `VkExternalMemoryHandleTypeFlagBits handleType`.
 - Added `VMA_VERSION` macro with library version number (#507).
 - Improvements in the algorithm choosing memory type when `VMA_MEMORY_USAGE_AUTO*` is used (#520).
-- Fixes for compatibility with C++20 modules on Clang 21 and GCC15 (#513, #514).
+- Fixed compatibility with C++20 modules on Clang 21 and GCC15 (#513, #514).
+- Fixed race condition in defragmentation (#529, #313).
 - Other fixes and improvements, including compatibility with various platforms and compilers, improvements in documentation, sample application, and tests.
 
 # 3.3.0 (2025-05-12)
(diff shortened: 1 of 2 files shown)


sawickiap:
Thank you for this fix. It looks good to me. I merged it into my branch for now:

PR #542, opened 2026-08-19
Fixes for compilation on 64-bit LoongArch

Cacodemon345:
This PR add a preprocessor check for the `__loongarch_lp64` defines in various places, which is defined when the compilers target the 64-bit LoongArch instruction set architecture.

Reference: https://github.com/loongson/la-toolchain-conventions/


sawickiap:
Thank you for this contribution. I merged it - to my fork for now:

